Issue #29, opened 2019-10-04
Wanderer prefs detailed settings do not save correctly on big endian systems.


wawatok:
including (tested on) m68k.
the problem must be in
https://github.com/aros-development-team/AROS/blob/master/workbench/prefs/wanderer/wpeditor.c
there is a number of macros responsible for this, among others SAVEVIEWSETTINGSTAG
as far as i know these settings should be asved in little endian for all platforms, so that the prefs files could be interchangeable.
so the numeric prefs should probably be endian converted for big endian platforms using aros det of macros like LONG2LE or something like that.

Issue #27, opened 2019-09-27
Distance Label From Image Icon with MagicWb Icon and ClassicWb Icon


Zilog80c:
Hi to all,
  problem there is both Native Mode and Rtg Mode  

  AROS-20190926-amiga-m68k nightbuild

  I try to use old 16 color MagicWb icon (but also classic wb icon), but the label of text are too distance from image icon ,such as the image in attached file.

I try to change setting into wanderer prefs, but the result non change.

Thanks a lot for help.

![Incon_Align](https://user-images.githubusercontent.com/31391699/65793312-cd128400-e165-11e9-8d69-1affb02a1bdc.JPG)

PR #33, opened 2019-10-20
Fixing problem with saving and loading Wanderer global.prefs file on BE systems

bsek:
This small change should hopefully fix issue #29 and #27


wawatok:
have tested this change in x86_64 and on m68k and saving prefs works with it as well as interchanging prefs files between these platforms.

michalsc (approved):

PR #34, opened 2019-10-28
Fixing redraw issue in iconlist.c if drawers/files are shown in list mode

bsek:
Added height of header row to the clip region used. Since this was missing, the upper row was not re-drawn when scrolling up earlier.


wawatok:
i confirm this fix working, tested on x86_64 linux hosted.

wawatok:
seems some debug is still on:
[IconList] IconList__MUIM_HandleEvent: IDCMP_NEWSIZE

PR #35, opened 2019-11-01
Centers titles vertically in screen and window title bars

bsek:
This small PR centers the text in the title bars vertically, when skinning is turned off, by adjusting the offsets. To me personally that looks better than having them aligned to the top, but I'm not sure if this change should be distro specific or if it belongs in the main repository.

### DIFF
--- a/rom/intuition/scrdecorclass.c
+++ b/rom/intuition/scrdecorclass.c
@@ -1,6 +1,6 @@
 /*
-    Copyright Â© 1995-2013, The AROS Development Team. All rights reserved.
-    Copyright Â© 2001-2003, The MorphOS Development Team. All Rights Reserved.
+    Copyright © 1995-2019, The AROS Development Team. All rights reserved.
+    Copyright © 2001-2003, The MorphOS Development Team. All Rights Reserved.
     $Id$
 */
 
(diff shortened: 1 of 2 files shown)


Kalamatee:
# If possible please try to confirm the "default" behavior against AmigaOS. If it is not the expected behavior, then perhaps it can be done in a (runtime) configurable way.
# It would be nice to not use hard coded values, and/or to have a comment about what the values are and why

wawatok:
the behaviour comparison against the amiga os is that the title text, as long as decoration is disabled  are not centred on aros while it is centred on amiga os. i have tested this and with bseks patch these offests looked fine again. with different fonts and sizes. since without decoration switched on, the height of title bar and window bar are hard coded id think its ok to use hard coded offset. again, it influences only the look without decoration.

Kalamatee:
As posted on slack -:

If you can make the following adjustments I will merge it after you have changed it...

Kalamatee  3 minutes ago
1> update the copyright notices in both the files you have touched at the top (preferably fixing the copyright symbol while at it ;D)

Kalamatee  2 minutes ago
2> dont use hard coded sizes but determine them using the provided structures and values -:

Kalamatee  1 minute ago
in srdecorclass.c you should probably be using ...

    Move(rp, msg->sdp_Screen->BarHBorder, msg->sdp_Screen->BarVBorder + rp->TxBaseline + ((msg->sdp_Screen->BarHeight - TxHeight) >> 1));

... so that the adjustment is correct depending on the used font and actual size of the bar area

Kalamatee  1 minute ago
likewise in windecorclass.c -:

Kalamatee  1 minute ago
    if (right - left > 6)
    {
        ULONG                   textlen, titlelen;
        struct TextFont         tf;
        struct TextExtent       te;
        tf = DRI(msg->wdp_Dri)->dri_Font;
        SetFont(rp, tf);
        titlelen = strlen(window->Title);
        textlen = TextFit(rp
                          , window->Title
                          , titlelen
                          , &te
                          , NULL
                          , 1
                          , right - left - 6
                          , window->BorderTop - 2);
        if (textlen)
        {
            left = left + 3;
            SetAPen(rp, pens[(window->Flags & WFLG_WINDOWACTIVE) ? FILLTEXTPEN : TEXTPEN]);
            Move(rp, left, tf->tf_Baseline + ((window->BorderTop - tf->tf_YSize) >> 1));
            Text(rp, window->Title, textlen);
        }
    }

PR #103, opened 2020-03-09
Send a handshake and set CIA serial in input mode during keyboard ini…

michalsc:
…tialization. This helps against keyboard missing after reset.


mattrust:
Reverted by 7b60d0625af2be8b6ee4da1049048fe1c764d51b because it was only for Vampire

PR #105, opened 2020-03-12
correct mouse sprite planar data in saga hidd.

wawatok:
change tested and submitted by apollo team.


wawatok:
will be updated and reissued.

PR #112, opened 2020-04-09
revert minor cleanup changes

wawatok:
as they actually broke arial and topaz fonts.
reverted commit: 2de072b975ec4122cbd45fd0a7bdc1d29cb2fed4


wawatok:
has been tested on amiga-m68k font prefs and ISO-ArialPL as well as ISO-topazPL display and can be chosen again in fonts prefs. the reverted change was meant for name and capitalization consistency. but since the font names are embedded in binaries, renaming font files and dirs renders them unusable. i dont see hex editing binaries as an option because the name sting length would differ. reverting is the safest option at this time imho.

PR #120, opened 2020-05-29
Fix copyright symbol to (C)

stanman29:
Change to just a pure text copyright symbol (C), which also silences the same warnings. This fixes my stupid mistakes.


Kalamatee:
Im not sure it is desirable to use (c) instead of the actual symbol in all cases, so some should be using "\xA9" (not 0xA9).

stanman29:
So should they all be reverted first? How do I test which one's need to be modified to hex \xA9? Is there a test?

Kalamatee:
I have applied a corrected version now so will close this request.

PR #178, opened 2020-12-18
Mingw32 fix

deadw00d:
Fixes to make mingw32 boot

### DIFF
--- a/arch/all-mingw32/bootstrap/memory.c
+++ b/arch/all-mingw32/bootstrap/memory.c
@@ -55,7 +55,7 @@ void *AllocateRAM32(size_t len)
             addr = NULL;
         }
     }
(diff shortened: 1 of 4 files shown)


Kalamatee:
I seem to recall the code being reverted by 0d4a9eb is because not all headers provide NtCurrentTeb.
It would be good if this can be confirmed, and if so a portable way to access it used instead.

deadw00d:
Hmm, I think I found also similar comments on Internet. I personally used Mingw64 under Linux MInt and it has those headers.

The issue is that with this code disabled, AROS does not even boot. It looks up in Enable.

Kalamatee:
Ive looked a little into it - and from what I can gather it seems to be headers for older versions of windows that lack it - atleast from Windows 7 onwards it _should_ be available.

PR #345, opened 2021-01-07
Update pictureclass.c

Miker-arosdev:
Test Pull Request to Update Picture Datatype.


Miker-arosdev:
Will be updated with new files.

PR #402, opened 2021-01-09
 Bugfixes and reworked file io dialog

bsek:
Bugfixes and reworked file io dialog

Fixed a few bugs related to copy and delete operations in Wanderer.
Changed the file io dialog to be less verbose, the old one caused slowdowns
on less powerfull systems. It will now display two progress bars; one
displaying which of the dropped items is being processed and one displaying
the status for the files being copied/deleted in that particular item.


bsek:
These changes requires this branch https://github.com/aros-translation-team/wanderer/tree/copyimprovements to be merged as well.

PR #556, opened 2021-03-23
Tiffdtype

Miker-arosdev:
Added TIFF Datatype to Classes/Datatypes. Added TIFF.dtd to Devs/Datatypes.


polluks (commented):
Missing big-endian support

PR #571, opened 2021-08-26
Don't default to 2 color mode (Black & White) during boot on Amiga

MBeijer:


MBeijer:
> Hi Marlon - it might be an idea to add a bit more verbosity about why this change is necessary.

Hi Nick,

Sorry, I thought it was verbose enough! :D On Amiga in earlyboot / boot with no startup-sequence everything is black and white. It works fine for early boot menu, but when you get into the "no startup-sequence" part, it should have more colors. This is at least how it is on Kickstart 3.1.

This change goes from 1bit mode (black and white) to 2bit mode (4 colors) to mimic the behaviour on classic Amiga kickstarts. Works fine on OCS, ECS and AGA.

Is this undesired?

PR #581, opened 2022-05-04
Safechanges

deadw00d:
This is a set of unrelated small changes which should generally be "safe". They were tested in my repository on linux-x86_64 and amiga-m68k targets, as well as build tested with linux-x86_64 in this repository.


Kalamatee:
Hi @deadw00d - 058adab causes a conflict (trunk use 2.12.1 now), and 6b66ac8 doesn't apply (we use version 1.2.12). Can you remove/adjust those and issue a new pull request please?

deadw00d:
Ok, I will fix the freetype2 patch. As for zlib patch, I didn't get conflict when rebasing. The change of download directory to /fossils/ should not cause problems, because that directory also contains the most recent release. The change was made so that when next release of zlib comes out, the build is not broken. If you prefer me to remove this change, let me know.

deadw00d:
Branch is now rebased and freetype conflict fixed.

PR #583, opened 2022-05-04
Time64

deadwood2:
This is a proposal of implementation of future 64-bit time support. What the changes do now is to freeze certain elements on 32-bit "forever" - essentially places where relative time is used and 32-bits is enough as well as document how the implementation can be done in future in a way that no longer breaks compatibility.

Note: side effect of these structure changes is that structure copying no longer works in some cases even though structures have same fields. This means field by field copying is needed.

### DIFF
--- a/developer/demos/simple_timer.c
+++ b/developer/demos/simple_timer.c
@@ -202,8 +202,8 @@ void wait_for_timer(struct timerequest *tr, struct timeval *tv )
 
 tr->tr_node.io_Command = TR_ADDREQUEST; /* add a new timer request */
 
-/* structure assignment */
-tr->tr_time = *tv;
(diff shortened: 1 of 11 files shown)


deadw00d:
Functionally tested in my repository on linux-x86_64 and amiga-m68k (for backwards compatibility). Build tested on linux-x86_64 and pc-x86_64 in this repository.

deadw00d:
FYI, the Codacy warnings are invalid in this case. We are setting fields of the request and passing the request to DoIO.

deadwood2:
Since these changes has not been accepted in total, how about I only limit this PR to changes that have no interface impact, so essentially changes that replace assigning whole timeval structure into assigning it's two fields separately?

PR #593, opened 2022-12-25
Misc changes 2022 H2

deadw00d:
Number of changes to different components. Build-tested on linux-x86_64


deadw00d:
One of the changes here seems to stop raspi-armhf from building. I will look into this and submit another PR.

PR #594, opened 2022-12-26
Place output of build in TOOLDIR

deadw00d:
TOOLDIR is specific to build, while CROSSTOOLSDIR can point to a shared, external crosscompiler. Placing output of one build in shared directory confuses other builds and causes them to fail. This PR needs to be merged together with matching PR in Contrib repository.


deadw00d:
I've made a mistake in this PR that causes the package to be searched in wrong place. I will submit a fixing PR.

PR #596, opened 2022-12-30
Fix build under ARM

deadw00d:
va_list is not a pointer, but a structure under ARM. Follow approach used in locale.library

### DIFF
--- a/workbench/libs/muimaster/mui_request.c
+++ b/workbench/libs/muimaster/mui_request.c
@@ -41,6 +41,16 @@ AROS_UFH2S(void, len_func,
     
     AROS_USERFUNC_EXIT
 }
+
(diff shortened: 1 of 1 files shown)


deadw00d:
FYI: there also seems to be some problem related to mingw32-x86_64 build and I can't figure it out - some sort of linking problem. Hopefully Kalamatee can look into this at some point.

deadw00d:
Also, maybe you can consider enabling discussions features on AROS repo. This would allow keeping all the discussions and resolving of problems related to my code merges into one thread. My general intention is do provide my work on a quarterly / half-year basis.

mattrust:
Regarding discussions: done. Please tell me if it works for you.

PR #621, opened 2023-06-28
fix Raspberry Pi support

NinjaCowboy:
I've ran into plenty of issues trying to get AROS running on a Raspberry Pi 3, so this ongoing PR attempts to fix a few of them.

In binutils, I disabled RELA relocations and use REL instead, since the addends all get ignored during the final link of an ELF file for some reason, breaking a lot of code (notably the PC-relative jumps like non-local "bl" which use these to adjust for the PC prefetch bias).

I am also compiling everything with `-mword-relocations`. arch/arm-raspi/boot/elf.c doesn't seem to compute R_ARM_MOVT_ABS relocations properly (frankly, I don't know exactly how these are supposed to work, either), and compiling with the `-mword-relocations` flag avoids generating these (uses pool loads instead). 

The boot process gets quite far, but there are a few `bad pointer` and `Software Failure!` errors generated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@
 /.externalToolBuilders
 /autom4te.cache
 /.vscode
+*.swp
 
 # /arch/.unmaintained/amiga/workbench/hidds/graphics/intuition/
 /arch/.unmaintained/amiga/workbench/hidds/graphics/intuition/*.bak
--- a/arch/arm-native/kernel/mmakefile.src
+++ b/arch/arm-native/kernel/mmakefile.src
@@ -76,6 +76,7 @@ $(OSGENDIR)/boot/core.elf: $(KOBJSDIR)/kernel_resource.o $(KOBJSDIR)/exec_librar
 		@$(ECHO) "Creating   $@"
 		%mkdirs_q $(OSGENDIR)/boot
 		@$(TARGET_LD) -Map $(OSGENDIR)/boot/core.map -T $(SRCDIR)/$(CURDIR)/ldscript.lds -o $@ $^ -L$(AROS_LIB) -larossupport -lautoinit -llibinit -lstdc.static -laeabi
+		$(TARGET_OBJCOPY) --only-keep-debug $@ $@.dbg
 		@$(TARGET_STRIP) --strip-unneeded -R .note -R .comment $@
 
 #MM kernel-kernel-raspi-arm : includes
--- a/arch/arm-native/kernel/platform_bcm2708.c
+++ b/arch/arm-native/kernel/platform_bcm2708.c
@@ -2,6 +2,8 @@
     Copyright (C) 2015-2016, The AROS Development Team. All rights reserved.
 */
 
+#undef __AROSEXEC_SMP
+
 #include <aros/types/spinlock_s.h>
 #include <aros/kernel.h>
 #include <aros/symbolsets.h>
@@ -92,7 +94,7 @@ static void bcm2708_init(APTR _kernelBase, APTR _sysBase)
         asm volatile ("mrc p15, 0, %0, c2, c0, 0":"=r"(tmp));
         ((uint32_t *)(trampoline_dst + trampoline_data_offset))[0] = tmp; // pde
         ((uint32_t *)(trampoline_dst + trampoline_data_offset))[1] = (uint32_t)cpu_Register;
-
+#if 0
         for (cpu = 1; cpu < 4; cpu ++)
         {
             cpu_stack = (uint32_t *)AllocMem(AROS_STACKSIZE*sizeof(uint32_t), MEMF_CLEAR); /* MEMF_PRIVATE */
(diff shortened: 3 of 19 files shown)


Kalamatee:
I haven't had time to fully look over this but a number of changes are incorrect/unnecessary/unwanted.

FWIW the arm build hasn't been working since Michal attempted to make it use proper elf files, but prior to that, it was able to boot. I have not had time to look into what was not reverted correctly, but that should've been all that was necessary to get it back to a booting state.

camthesaxman:
When was the last commit where the ARM version was able to boot correctly? I may be able to fix the issues if I know what actually works.

Kalamatee:
I'm not sure exactly, and would need to test to find out - but it should be in the 2019 timeframe.

NinjaCowboy:
Thanks, guys. Ah, the new elf support probably explains why I'm having these reloc issues, so I'll try building some old commits. What was it using before elf files? AROS has used elf for as far as I can remember.

NinjaCowboy:
I tried to build commit c7205827715ae92a4952fe92c4cea3a12c8df866 from 17 March 2019, since this seems to be the last one that dealt with ARM before adding ELF support, but I get a lot of these kinds of errors.
`/home/ninja/programming/aros-rpi/AROS-old-build/bin/linux-x86_64/tools/crosstools/arm-aros-ld: error: /tmp/catmpcYxbGu uses VFP register arguments, /home/ninja/programming/aros-rpi/AROS-old-build/bin/raspi-arm/AROS/Developer/lib/libposixc.a(posixc_fprintf_stub.o) does not`

NinjaCowboy:
These "bad pointer" errors that I am getting like `rom/oop/setattrs.c, 59: bad pointer: attrList = $f845134c` come from the ASSERT_VALID_PTR macro. ASSERT_VALID_PTR calls TypeOfMem, which then calls FindMem to look up an address in `SysBase->MemList`. However, this check never succeeds, even with what should be valid memory. I've called `debugmem()` to inspect `SysBase->MemList`, but it only shows one memory region `List System Memory: Attr=00009505 from 0x0005d000 to 0x3bb00000 Free=1000273536`, and those appear to be physical addresses (not virtual). What exactly is supposed to be adding entries to `SysBase->MemList` ? It never gets populated when a new library is loaded.

Kalamatee:
> I tried to build commit [c720582](https://github.com/aros-development-team/AROS/commit/c7205827715ae92a4952fe92c4cea3a12c8df866) from 17 March 2019, since this seems to be the last one that dealt with ARM before adding ELF support, but I get a lot of these kinds of errors. `/home/ninja/programming/aros-rpi/AROS-old-build/bin/linux-x86_64/tools/crosstools/arm-aros-ld: error: /tmp/catmpcYxbGu uses VFP register arguments, /home/ninja/programming/aros-rpi/AROS-old-build/bin/raspi-arm/AROS/Developer/lib/libposixc.a(posixc_fprintf_stub.o) does not`
> 

Yes - we had issues getting the toolchain built correctly (the toolchain is generally built to be generic for AROS targets, using the correct flags for specific targets at compile time).

> I'm aware that a lot of the changes I made in this PR are unwanted. I will clean that up later.
> 
> I keep running into endless build errors trying to compile old versions of AROS, so I'm going to give up on trying to get that to compile, and just work from the current version instead.

Ok - it was meant more to look at what changed in regards to the RasPI target, causing it to break in the reported ways.

Kalamatee:
> Thanks, guys. Ah, the new elf support probably explains why I'm having these reloc issues, so I'll try building some old commits. What was it using before elf files? AROS has used elf for as far as I can remember.

To clarify - AROS uses "Elf" format, but not real executables. We had discussed using real executables, which is what Michal was attempting to implement for ARM, but eventually he gave up and instead started working on the arm-m68k emulator that takes up his time these days. So the ARM build is in a broken state due to this - its somewhere between the old working build, and using proper executable elf binaries.

Kalamatee:
> These "bad pointer" errors that I am getting like `rom/oop/setattrs.c, 59: bad pointer: attrList = $f845134c` come from the ASSERT_VALID_PTR macro. ASSERT_VALID_PTR calls TypeOfMem, which then calls FindMem to look up an address in `SysBase->MemList`. However, this check never succeeds, even with what should be valid memory. I've called `debugmem()` to inspect `SysBase->MemList`, but it only shows one memory region `List System Memory: Attr=00009505 from 0x0005d000 to 0x3bb00000 Free=1000273536`, and those appear to be physical addresses (not virtual). What exactly is supposed to be adding entries to `SysBase->MemList` ? It never gets populated when a new library is loaded.

It fails because the kernel has been mapped into virtual memory, but no memory region has been registered to express this, so the asserts correctly declare it is invalid. The kernel should register a region matching the used mappings.

Kalamatee:
p.s if you have frequent questions, you would be better asking on the linen/slack channel(s) ;)

NinjaCowboy:
Thanks. Do you have a link to the Slack channel?

Kalamatee:
> Thanks. Do you have a link to the Slack channel?

ss https://github.com/aros-development-team/AROS/wiki

PR #622, opened 2023-07-17
Fixes 02

deadw00d:
Two fixes:
a) Prefs/Pointer was not longer saving selected pointer (after recent compatibility changes to muimaster.library)
b) URL for acpica has changed


deadw00d:
I didn't rename the existing function, though I added a third function.

Issue #630, opened 2023-11-09
Wanderer, Rename File or folder lower case


AMIGASYSTEM:
On Wanderer on it is possible to rename File or Folder by changing from lower case to upper case or vice versa.

This only happens if you rename "using the same name", if you change only one letter or all the letters. 

No problem, however, if you rename with a different name.

From Shell and from Dopus no problem renaming from lower case to upper case or vice versa.

https://www.arosworld.org/infusions/forum/viewthread.php?thread_id=1142&getfiles=109

PR #633, opened 2024-02-02
Misc fixes

deadwood2:
Set of different fixes. Please check each commit for description and rationale.

Build-tested with linux-x86_64


deadwood2:
FYI: some of these commits fix https://github.com/aros-development-team/AROS/issues/630

PR #637, opened 2024-02-10
Make AROS CDVDFS work on AmigaOS (again)

reinauer:
- GFLG_GADGHIGHBITS: Fix incorrect usages.
- rom/filesys/afs: Fix issue where OpenFromLock() returns NULL on success
- Fix for adding windows entry to grub
- Adjust to change in keymap limiting Hi to 0x38
- Further relax condition
- Avoid generating empty .fd files as they have no value
- Also filter out devices without functions
- Allow renaming into exclusively locked directory
- Pass the tags containing project name and lock to CLI launch
- Set the correct mask via ohciEnableInt when enabling interrupt
- After MUIA_String_Contents, MUIA_String_Acknowledge has same value
- 2023 -> 2024
- Increase priority of close function to guarantee free() working
- Extend check during rename to allow change of capitalization of name
- Don't translate error when it is already DOS error
- Allow changing capitalization of name when renaming
- Allow changing capitalization of name when renaming
- Fix warnings
- 64-bit fix.
- Fix reading random memory
- 64-bit fix
- Silence debug
- Listtree.mcc: Rework CreateImage workaround for 64-bits
- Fix warnings related to C library functions
- Add workaround for problem visible in C++ code
- Extend CreateImage hack
- Fix for '*' in sscanf
- Port newer version of random.c from FreeBSD
- Fix: Mark ata_32bit object for '32bit' argument
- Replace enableusb option with noioapic
- Add information at the top to avoid confusion
- Default state of ata_32bit is disabled
- Add separate tabs for ATA, AHCI and NVME
- Implement support for ATA=disable argument
- Add support for AHCI=disable and NVME=disable
- Implement AHCI=force150/force300/force600
- Localize strings
- Update version
- Update ftmanager.c
- Catalog submodules updated.
- Use github mirror of PCIIDS project. Main site sometimes fails.
- CDVDFS: Trim whitespace
- CDVDFS: Use CopyMem where possible
- CDVDFS: Fix dbprintf / dbinit / dbuninit
- CDVDFS: Allow use of OpenLibrary for opening libraries
- CDVDFS: Ensure filesystem IDs are available
- CDVDFS: Use %lx instead of %p for pointers
- CDVDFS: Get CDVDFS working on AmigaOS
- CDVDFS: Fix manual dependencies in GNUMakefile
- CDVDFS: Serial Debugging on AmigaOS
- CDVDFS: Properly cast parameters for FindToolType()
- CDVDFS: Cleanup_CDROM: Don't dereference p_cd if it is NULL
- CDVDFS: checkcd: Fix inverted NULL pointer check
- CDVDFS: cdrom.c: Check pointers in Start_Play_Audio() and Stop_Play_Audio()
- CDVDFS: device.c: Check return code of Mount_Check()
- CDVDFS: devsupp: Secure string copies in Get_Startup()
- CDVDFS: generic.c: Don't dereference potential NULL pointers
- CDVDFS: hfs: file names can be 30 characters long
- CDVDFS: generic: Prevent NULL pointer dereference in Close_Object()
- CDVDFS: Use D(bug()); consistently where applicable
- CDVDFS: Bring version numbers and date in sync


reinauer:
Woah not sure what happened here

PR #645, opened 2024-07-21
Fix crash during boot of native compile with --enable-debug

thatguywiththekids:
The WIP resource log.resource made AROS crash during boot if AROS was compiled with --enable-debug. This pull request makes logAddEntryA refuse to add the log entry instead of attempting to perform I/O on a device null pointer. It also fixes some memory management errors.

log.resource does not perform any actual log broadcasting currently anyway, so I consider this a reasonable fix for now.


thatguywiththekids:
Is there an issue with this?

mattrust:
Merged without testing. I don't know why Kalamatee didn't check.

PR #683, opened 2025-04-21
Quad pthread ahi

deadwood2:
This PR contains three functional themes in consecutive order:

1) Change to LCQUAD macros to indicate type they operate in
2) Updates and fixes to pthread library
3) Restoring source compatibility to AHI

build-tested with linux-x86_64


deadwood2:
I reviewed two issues reported by Codacy. On the stacksize68k, it is actually #ifdefed for __MORPHOS__ only so codacy should not be analyzing it in my opinion. For the variable inf, isn't it actually Amiga/AROS code-style to follow the older C specification and declare all variables on beginning of the block? I think that's how majority of our codes is written :/

So to sum up, I think both of those warning should be ignored.

PR #691, opened 2025-07-19
Fix a little regression introduced in 4b5ac577c5be33ac7007ebf74f60fc5d914c5ec2

thatguywiththekids:


Kalamatee:
I'm not convinced this is correct. It is documented that the alignment of memory returned by AllocPooled() is determined by the memFlags passed to CreatePool()

If you use MEMF_ANY or MEMF_PUBLIC, the alignment is typically 4 bytes.

If you use MEMF_CLEAR | MEMF_PUBLIC, it’s still 4 bytes.

If you use MEMF_32BIT or MEMF_FAST, alignment could be 32-bit or more, depending on the CPU and OS version.

Also please don't use // for multi-line comments - try to keep it consistent with the existing code.

thatguywiththekids:
That's the complexity I was hoping we could avoid :)

thatguywiththekids:
I see this was merged, but I did some followup tests anyway. I have tested 3 byte allocations from 5 byte puddles for ANY, CHIP, FAST, 24BITDMA, PUBLIC, and I have never seen four byte alignment that wasn't also eight byte alignment.

PR #700, opened 2025-09-06
AllocPooled adjustments

thatguywiththekids:
AMIGA define is not exclusive to the Amiga target. Just set the same offset for all targets willingly, and dial it down to 8 bytes which is what we have seen on AmigaOS. No pointer is larger than 8 bytes, unless AROS should run on IBM Teraspace systems and then that's a whole project :)

### DIFF
--- a/rom/exec/createpool.c
+++ b/rom/exec/createpool.c
@@ -88,14 +88,14 @@
 
     /*
      * puddleSize needs to include MEMHEADER_TOTAL, allocator context size and
-     * ALLOCPOOLED_USER_OFFSET.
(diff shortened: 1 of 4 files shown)


Kalamatee:
I would probably just make it use AROS_WORSTALIGN, but need to try and make time to think on it.

thatguywiththekids:
AROS_WORSTALIGN is 4 on Amiga, so it would break AmiBlitz to use it for AllocPooled (need 8 byte alignment for the AGA fetch modes used).

thatguywiththekids:
Good point about SSE. I'll look into adjusting the PR.

PR #719, opened 2025-10-29
rexxsyslib: CreateArgstring and CreateRexxMsg should return value in both A0 and D0 on Amiga

thatguywiththekids:
Autodocs say so and I have confirmed the behaviour for CreateArgstring on AmigaOS 3.1.

The return value is copied to A0 for all __mc68000__ targets for a simpler condition.


thatguywiththekids:
Updated from feedback

Issue #709, opened 2025-10-09
local builds issue with libaom


agreppin:
Hello,

I'm trying to make a local nightly build of AROS.
That always fail with `cannot find -laom`.

I'm using those command lines from the nightly build logs:

```log
configure --target=pc-x86_64 --enable-target-variant=smp \
--with-gcc-version=15.2.0 --enable-ccache --enable-build-type=nightly \
--with-portssources=/a/src/AROS/bin/Sources \
--with-aros-toolchain-install=/a/x-tools/x86_64-aros
```
... idem with

```log
configure --target=linux-x86_64 --with-gcc-version=13.4.0 --enable-ccache \
--enable-unit-tests=no --enable-build-type=nightly \
--with-portssources=/a/src/AROS/bin/Sources \
--with-aros-toolchain-install=/a/x-tools/aros/linux-x86_64-13.4.0

MMAKE] Making workbench-datatypes-html-fd in workbench/classes/datatypes/html
Generating AROS/Developer/SDK/fd/html_lib.fd
[MMAKE] Making workbench-datatypes-pnm-fd in workbench/classes/datatypes/pnm
Generating AROS/Developer/SDK/fd/pnm_lib.fd
make[2]: *** No targets specified and no makefile found. Stop.
make[2]: *** No targets specified and no makefile found. Stop.
[MMAKE] Making workbench-datatypes-heic in workbench/classes/datatypes/heic
Makedepend workbench/classes/datatypes/heic/heicdummy.cc ...
Generating support files for module heic
Makedepend bin/linux-x86_64/gen/workbench/classes/datatypes/heic/heic/heic_end.c ...
Makedepend bin/linux-x86_64/gen/workbench/classes/datatypes/heic/heic/heic_start.c ...
Makedepend workbench/classes/datatypes/heic/heicclass.c ...
Compiling bin/linux-x86_64/gen/workbench/classes/datatypes/heic/heic/heic_start.c 
Compiling workbench/classes/datatypes/heic/heicclass.c 
Compiling workbench/classes/datatypes/heic/heicdummy.cc 
Compiling bin/linux-x86_64/gen/workbench/classes/datatypes/heic/heic/heic_end.c 
Building Module AROS/Classes/DataTypes/heic.datatype ...
[...]

PR #720, opened 2025-11-02
revert PR #710, fixes issue #709

agreppin:
LLVM toolchain build ok, problems later with duplicate symbols SysBase/DOSBase


agreppin:
more info on this issue: I made a log of all the objects that are built after those 3 steps
- install-LLVM-stripped
- install-lld-stripped
- install-clang-resource-headers-stripped

seems that the same ~2000 objects are rebuilt over and over again.
[llvm-objs-list.zip](https://github.com/user-attachments/files/24085061/llvm-objs-list.zip)

PR #728, opened 2026-01-01
Make the mouse pointer visible again when leaving the "nomedia" animation

thatguywiththekids:
Changes also include some simplifications of the source code: __OOP_NOATTRBASES__ and __OOP_NOMETHODBASES__ are always set, so the alternate code path (which also seems to have some errors) can be removed.


Kalamatee:
We don't want the mouse pointer visible until a display "opens" - so I'm not convinced this change is wanted.
I don't really see the need for the dosboot changes either - they prevent it being compiled without __OOP_NOATTRBASES__, for testing purposes. It doesn't really serve any purpose/fix anything or introduce anything new.

PR #733, opened 2026-01-15
append full filename to panel statusbar in wanderer

dreamos82:
This pull request is to try to fix lack of full filename visibility on wanderer. 

When file/folder name  is longer than 10 characters, it is truncated, and there is no way to see its full name, neither on list view nor on icon view. 

This simple PR is trying to partially fix this issue. 

When a file/folder is clicked on the status bar the string: "Selected: fullfilename" is appended to the already existing one. 

The case covered are only when mouse is clicked: 
* no "selected" string and file is selected  -> nme of selectred file  appended the filename
* "selected" is present and file is selected -> THe current filename  is replaced with the new name
* clicked on an area where there are no files (causing current selections to be cleared) -> Selected string is removed
* Multiple files are being selected (with shift pressed) -> Only the last filename is appended.

This cover only the mouse click events. So far i was unable to find/understand how the slection via keyboard is fired in wanderer (if anyone knows it i can try to extend this code also to the keyboard case). 

This is a screenshot of the output: 
image

Another thing is that to be able to access the panel_Statusbar objects, I had to move  the `panel_StatusBar_DATA` struct  outside the `panel_statusbar.c` file and create an header for it (`panel_statusbar.h`) 

I hope this doesn't break any coding rule (or any logic inside the Wanderer Code, but the main problem is that i haven't found another way to access the statusbar from the iconwindov events., if there is a better way please let me kwno.

### DIFF
--- a/workbench/system/Wanderer/iconwindow.c
+++ b/workbench/system/Wanderer/iconwindow.c
@@ -54,6 +54,8 @@
 #include "iconwindow_iconlist.h"
 #include "iconwindowbackfill.h"
 
+#include "panel_statusbar.h"
+
 #include "version.h"
 
 #if defined(ICONWINDOW_NODETAILVIEWCLASS)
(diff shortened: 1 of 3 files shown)


dreamos82:
Btw although yeah a tooltip box can be used (if you can point me where it is i can try to move the code there), anyway many file managaers (like KDE Dolphin) shows also the full name on the status bar,, so this is one way of addressing the issue, and was meant only an initial implementation, and the easiest to implement.

PR #768, opened 2026-05-05
misc: ExAll padding, cdceth packing, Startup-Sequence ordering

bsek:
- pfs3, ram, diskfont. pad ExAll entry size to AROS_PTRALIGN instead of 2 bytes so the next entry starts at a properly aligned address on platforms where pointers need stricter alignment than 16 bits.
- cdceth. drop the global #pragma pack(2) from the class header. it applied to all included system structures and produced misaligned accesses on ARM. set DEBUG=0 in mmakefile.
- Startup-Sequence. run FailAt before SetClock so a SetClock failure does not abort the boot.

### DIFF
--- a/rom/filesys/pfs3/fs/directory.c
+++ b/rom/filesys/pfs3/fs/directory.c
@@ -1209,7 +1209,7 @@ static ULONG FillInData(struct ExAllData *buffer, LONG type,
 	}
 
 	/* check fit */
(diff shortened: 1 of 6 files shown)


Kalamatee:
please split this up into 3 separate pull requests.

bsek:
ok, will do

PR #772, opened 2026-05-07
Shell: Don't override cli_Background if already set

metaneutrons:
A shell launched with cli_Background=DOSTRUE (e.g. via RunCommand from a script) should not have its background state overridden by the IsInteractive() check.


metaneutrons:
Closing — Copilot review correctly identified that this change would prevent interactive shells from returning to interactive mode after running a script. The original behavior is correct.

PR #774, opened 2026-05-07
kernel/tlsf: Add double-free detection

metaneutrons:
Check if a block is already free before freeing again. Prevents TLSF free-list corruption. Important on SMP where race conditions can cause double-frees.


metaneutrons:
Closing — will resubmit after Copilot review on fork. Fixed: moved double-free check after semaphore acquire for SMP safety.

PR #779, opened 2026-05-07
exec: Fix SMP race in OpenLibrary/OpenDevice

metaneutrons:
Use EXEC_UNLOCK_LIST_AND_PERMIT instead of EXEC_UNLOCK_LIST to ensure Permit() is called after releasing the list lock. Prevents deadlocks on SMP systems (x86_64-smp, aarch64-smp).

### DIFF
--- a/rom/exec/opendevice.c
+++ b/rom/exec/opendevice.c
@@ -97,7 +97,7 @@
 
     iORequest->io_Device = (struct Device *)FindName(&SysBase->DeviceList, devName);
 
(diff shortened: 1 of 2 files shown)


metaneutrons:
Closing — Copilot correctly identified that EXEC_UNLOCK_LIST_AND_PERMIT already calls Permit(), making the existing Permit() at the end of the function a double-Permit that corrupts TDNestCnt. The original code (EXEC_UNLOCK_LIST + later Permit) is correct.

PR #798, opened 2026-05-14
spinlock: gate 128-byte alignment by arch capability

bsek:
## Summary

Clang trusts the alignment hint and uses NEON instructions to access the struct field, which require their addresses
to be properly aligned. AROS's allocator only guarantees 16-byte alignment, so the attribute is effectively a promise the system can't keep. The result is a Data Abort the first time anything touches a `MsgPort`, `SignalSemaphore`, or any other struct that embeds a `spinlock_t`.

## What this PR does

Replace the unconditional `__attribute__((aligned(128)))` with a per-architecture macro `AROS_SPINLOCK_ALIGN`:

- `arch/i386-all/include/aros/cpu.h` defines it as `aligned(128)`
- `arch/x86_64-all/include/aros/cpu.h` defines it as `aligned(128)`
- All other architectures (ARM, etc.) get the default empty definition

### DIFF
--- a/arch/i386-all/include/aros/cpu.h
+++ b/arch/i386-all/include/aros/cpu.h
@@ -2,7 +2,7 @@
 #define AROS_I386_CPU_H
 
 /*
-    Copyright © 1995-2020, The AROS Development Team. All rights reserved.
+    Copyright (C) 1995-2020, The AROS Development Team. All rights reserved.
     $Id$
 
     NOTE: This file must compile *without* any other header !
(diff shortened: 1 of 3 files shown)


Kalamatee (approved):

stevereaver:
## Build error introduced by this PR

**File:** `workbench/classes/reaction/slider/sliderclass.c:34`

**Cause:** [sliderclass.c](cci:7://file:///home/reaver/aros-development/aros-source/workbench/classes/reaction/slider/sliderclass.c:0:0-0:0) defines `VOID_FUNC` as `void (*)(void)` guarded by `#ifndef VOID_FUNC`. Since `VOID_FUNC` is a typedef (not a macro), the `#ifndef` check always passes even after `libcall.h` has already defined it as `void (*)()`. This results in two conflicting typedef declarations.

```
error: redefinition of typedef 'VOID_FUNC' with different type
   34 | typedef void (*VOID_FUNC)(void);
      |                ^~~~~~~~~
In file included from .../include/clib/exec_protos.h:9,
                 from .../include/proto/exec.h:12,
                 from workbench/classes/reaction/slider/sliderclass.c:8:
.../include/aros/libcall.h:77: note: previous declaration of 'VOID_FUNC' was here
   77 | typedef void (*VOID_FUNC)();
      |                ^~~~~~~~~
```

**Fix:** Use the same guard macro as `libcall.h` and match its signature:

```c
// Before
#ifndef VOID_FUNC
typedef void (*VOID_FUNC)(void);
#endif

// After
#ifndef __typedef_VOID_FUNC
#define __typedef_VOID_FUNC
typedef void (*VOID_FUNC)();
#endif
```

PR #799, opened 2026-05-15
raspi: load USB classes from disk

bsek:
- Moved usb classes from rom to disk 
- Trimmed the usbromstartup to just register usb2otg driver

Device binding now happens during execution of startup-sequence.


Kalamatee:
This change prevents older PI's from starting boot on sdcard, but continuing from a USB device.

bsek:
> This change prevents older PI's from starting boot on sdcard, but continuing from a USB device.

Good point — I've put hub.class and massstorage.class back into the ROM so usb-msd boot still works

PR #801, opened 2026-05-15
dos: strip volume specifier in getpacketinfo

bsek:
Strip the volume/assign specifier before handing the name to the filesystem handler. GetDeviceProc() resolves the prefix to the correct handler + base lock, and the handler expects the remaining relative path - not the original name with a ':' still in it. 

Without this, handlers whose path parser treats ':' as a "reset to root" marker (at least the fat handler) end up searching the wrong directory and returning OBJECT_NOT_FOUND.

### DIFF
--- a/rom/dos/packethelper.c
+++ b/rom/dos/packethelper.c
@@ -37,7 +37,9 @@ BOOL getpacketinfo(struct DosLibrary *DOSBase, CONST_STRPTR name, struct PacketH
         phs->name = bstrname;
         return TRUE;
     } else { /* ":" */
-        BSTR bstrname = C2BSTR(name);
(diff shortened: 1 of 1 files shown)


Kalamatee:
Need to confirm if this behaviour is correct against amigaos.

deadwood2:
If that's of any help, I created a "replacement" header which allows compiling cunit tests for AmigaOS. This way the same test can be added to AROS and be used to verify behavior with AmigaOS:

https://github.com/deadwood2/AROS/blob/master/developer/debug/test/test-util.h

bsek:
Moved handing of specifier to fat.handler, closing this PR

PR #826, opened 2026-07-16
nlist: fix 64-bit taglist crashes (int values in stack-spilled varargs)

jonx:
On 64-bit targets, a plain C int passed as a taglist tag or value in a variadic call is only stored as 32 bits when the argument spills to a stack vararg slot (the first 8 arguments travel in registers on AArch64 and are zero-extended for free, which is why small tag lists never trigger this). The tag walker then reads the full 64-bit slot, and the stale upper half of the stack turns a harmless 0 into a garbage pointer such as 0x100000000.

Three call sites in the NList classes hit this in practice:

- `NListviews.mcp`'s `String2(0,80)` macro passes int 0 as `MUIA_String_Contents`; `String__OM_NEW` strcmp's the garbage pointer. This crashes Zune prefs ("GUI Settings..." in Wanderer) as soon as the NListviews settings page loads.
- `NList_mcc.c` and `NListview.c` OM_NEW pass `NoFrame` (= `MUIA_Frame, MUIV_Frame_None`) past the 8th vararg of `DoSuperNew`; frame 0 becomes 0x100000000, which `Area--MUIM_Setup` treats as a frame-spec string and strlen's. This crashes SysExplorer at launch.

Fix: cast the affected tags and values to IPTR at the call sites. In the NList_mcc site the spilled tag words themselves (`MUIA_Frame`, `Child` are 32-bit unsigned int constants) are cast too, since a garbage upper half there makes the walker silently skip the tag.

Found and verified on aarch64 (hosted): before the fix, both Zune WANDERER and SysExplorer crashed deterministically at the exact same instructions; after it, both open and run cleanly. x86_64 is exposed to the same C-semantics issue since varargs spill to the stack after 6 integer registers there.

### DIFF
--- a/workbench/classes/zune/nlist/nlist_mcc/NList_mcc.c
+++ b/workbench/classes/zune/nlist/nlist_mcc/NList_mcc.c
@@ -258,12 +258,15 @@ void obtain_pen(Object *obj, IPTR *pen, struct MUI_PenSpec *ps)
 
 #if !defined(__MORPHOS__)
 #ifdef __AROS__
-static __attribute__ ((noinline)) Object * VARARGS68K DoSuperNew(struct IClass *cl, Object *obj, Tag tag1, ...)
-{
-    AROS_SLOWSTACKTAGS_PRE_AS(tag1, Object *)
-    retval = (Object *)DoSuperMethod(cl, obj, OM_NEW, AROS_SLOWSTACKTAGS_ARG(tag1), NULL);
-    AROS_SLOWSTACKTAGS_POST
-}
+/* varargs cannot carry int-typed tag data on 64-bit targets; expand the
+   tag list into a full-width IPTR array at the call site instead */
+#include <aros/preprocessor/variadic/cast2iptr.hpp>
+#define DoSuperNew(cl, obj, ...)                                          \
+({                                                                        \
+    IPTR __args[] = { AROS_PP_VARIADIC_CAST2IPTR(__VA_ARGS__) };          \
+    struct opSet __ops = { OM_NEW, (struct TagItem *)__args, NULL };      \
+    (Object *)DoSuperMethodA((cl), (obj), (Msg)&__ops);                   \
+})
 #else
 static Object * VARARGS68K DoSuperNew(struct IClass *cl, Object *obj, ...)
 {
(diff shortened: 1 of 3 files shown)


jonx:
Fair points, reworked. The branch now has a single clean commit with no call-site casts at all:

- The two OM_NEW superclass attr lists are built as real `struct TagItem` arrays and passed via `DoSuperMethodA`. Aggregate initialization converts int-typed `ti_Data` to full width on any ABI, so the tag lists read naturally and nothing needs casting except genuine pointers. The modules' local varargs `DoSuperNew` helpers become unused and are removed.
- `String2` keeps one cast, `(STRPTR)(contents)`, inside the macro: that value is a pointer, so this is the same rule as passing `(char *)NULL` as an `execl()` sentinel.

On the tags you are right, they never needed casting: `GetTagsFromStack()` truncates the 64-bit read into the 32-bit `ti_Tag`, so a stale upper half is harmless there. Dropped.

On casting at the consumer instead: that cannot work for pointer-valued tags, because the information is destroyed at the call boundary. With a true varargs callee, `GetTagsFromStack()` reads every slot with `va_arg(ap, IPTR)` into `ti_Data`; when the caller passed a 32-bit int the upper half of the stack slot is simply garbage, and the consumer cannot distinguish "int 0 plus stack noise" from a legitimate 64-bit pointer (for MUIA_Frame the whole contract is small int = preset, otherwise pointer to a spec string). The reason practically no other tag usage in AROS needs casts is that it goes through the inline-stdarg variadic macros, which already convert every argument to IPTR; these classes opt out of those (`MUIMASTER_NO_INLINE_STDARG` is the muimaster default) and use the true-varargs linklib stubs, and in that configuration the call site is the only place with the type information.

Re-verified on aarch64 hosted after the rework: Zune WANDERER and SysExplorer both open and run with no traps.

jonx:
Done that way, and it is indeed nicer. Updated to a single commit:

- `NListview.c`: `#define DoSuperNew(cl, obj, ...) DoSuperNewTags((cl), (obj), NULL, __VA_ARGS__)` on `__AROS__`, the same line `Wanderer/iconwindow.c` already uses. Call site is back to a plain uncast tag list.
- `NList_mcc.c`: same macro with the `DoSuperNewTags()` body from `clib/alib_protos.h` spelled out locally, because that file builds with `NO_INLINE_STDARG` which compiles the SDK macro out.
- The nested `VGroup`/`NewObject ... End` children are created up front and passed as `Child, vgroup` / `Child, grp`: a builder's closing parenthesis lives inside the `End` macro, so it is invisible to macro argument collection and the nested form cannot sit inside the macro's arguments.

Two factual notes:

On `AROS_SLOWSTACKTAGS_ARG`: on the targets where this matters (`AROS_SLOWSTACKTAGS` platforms, aarch64/x86_64), the trio expands to `GetTagsFromStack()` and `_ARG(arg)` is just `tags`, the array it built; the `((IPTR)(x))` definition is the other variant. Either way the cast sits in the callee, and the damage happens in the caller: an int argument spilled to a stack vararg slot is a 32-bit store, and `GetTagsFromStack()` reads the slot back with `va_arg(ap, IPTR)` into `ti_Data`, upper half included. No callee-side cast can recover that.

On "the rest of the changes are then unnecessary": true for the DoSuperNew sites, but the Zune-prefs crash is a different call, `String2()` expands into a `MUI_NewObject()` linklib-stub call (muimaster's variadic macros are off by default via `MUIMASTER_NO_INLINE_STDARG`), so no DoSuperNew macro reaches it. The `(STRPTR)` cast on the contents argument inside that macro stays; it is a pointer-typed value in a true varargs call, same rule as the `(char *)NULL` execl sentinel.

Re-verified after the rework on aarch64 hosted: Zune WANDERER and SysExplorer both open and run with no traps.

jonx:
Trimmed as requested: comments reduced to a short neutral note on the 64-bit varargs constraint, no references to other components. The generated modules are byte-identical to the previously verified build, so behavior is unchanged.

Kalamatee (approved):

PR #846, opened 2026-07-22
aarch64: implement the CPU support layer

jonx:
This fills in the dormant arch/aarch64-all skeleton with a working AArch64 architecture layer, taken from a functioning AArch64 hosted port:

* exec: PrepareContext, StackSwap/NewStackSwap, CacheClearE, and register-transparent stubs for Disable/Enable/Forbid/Permit
* ExceptionContext with the full architectural register set (x0-x28, fp, lr, sp, pc, cpsr) and an optional FPU context pointer
* genmodule: AArch64 libcall stub/alias macros (aros/genmodule.h), hidden visibility for the per-module base accessors
* kernel: KrnBacktraceFromFrame via the AAPCS64 frame record
* stdc: setjmp/longjmp, fenv/fpmath dispatch, 128-bit long double in the static math set, plus the missing aros/types/fenv_t.h shared type header referenced by fenv.c and the stub fenv.h
* ELF: R_AARCH64_* relocations in dos/elf.h, InternalLoadSeg and the bootstrap elfloader, with range and alignment checks on JUMP26/CALL26 so unencodable branch targets are refused instead of silently truncated
* build system: aarch64 codegen/link specifics (large code model, --allow-multiple-definition including a KOBJ_LDFLAGS hook so the flag reaches the raw kobj link)

Notes for review:

* No build target selects aarch64 yet, so this is inert for existing targets. A hosted target that uses this layer follows in a separate PR; keeping the two apart should make both easier to review.
* Builds with the LLVM/clang 20.1.0 toolchain (--with-llvm-version=20.1.0) and stock lld using the standard "aarch64elf" emulation, so no lld patch is needed (in line with d8c49ed0ba, which builds stock lld for v20+).
* The first commit (collect-aros @response-file expansion, clang 16+ warning demotions) is not aarch64-specific, but both are prerequisites for building with a current clang.


jonx:
Adjusted: MODERN_CLANG_NOWARN now uses the NOWARN_ variables exposed from compiler.cfg. There was no probe yet for -Wincompatible-function-pointer-types, so I added one to config/features.in (features regenerated with the same 2.71 layout) and the WARN_/NOWARN_ pair to compiler.cfg.in, next to the existing incompatible-pointer-types entries.

jonx:
Moved: -mcmodel=large now comes from the top-level configure like the x86_64 one, as ISA_AARCH64_FLAGS in the base CPU ISA switch (with --target=aarch64-unknown-aros for llvm toolchains), emitted through aros_isa_extra. The base switch pattern also accepts arm64 triple spellings. make.cfg.in keeps only the aarch64 link flags.

Kalamatee (approved):

PR #860, opened 2026-07-24
dos: add an AArch64 GOT to the ELF loader

bsek:
ADR_GOT_PAGE/LD64_GOT_LO12_NC appear in every disk library through the genmodule symbols and C++ COMDAT groups, and need a GOT entry holding the symbol address. Entries live in hunk-backed blocks linked into the seglist, so UnLoadSeg frees them.

Also add the ADR_PREL_LO21, ADR_PREL_PG_HI21_NC, LD_PREL_LO19, CONDBR19 and TSTBR14 relocations, mirroring the boot loader in arch/aarch64-raspi.


Kalamatee:
the PC-relative relocation additions are welcome; the GOT belongs to the compiler contract, not the loader - AROS resolves every symbol at LoadSeg time, and the fix for GOT relocations is to configure codegen so they are never produced, exactly as the x86_64 port already does with -mcmodel=large + non-PIC

PR #864, opened 2026-07-26
m68k: reduce scheduler, interrupt and timer latency

warpdesign:
Improve scheduling latency on classic m68k Amiga systems, particularly under frequent serial interrupts.

## Main changes

  - Add faster m68k task-switch and dispatch paths.
  - Accelerate common Wait() and Signal() wake-up paths.
  - Reduce time spent with interrupts masked.
  - Shorten higher-priority task wake-ups.
  - Reduce fixed interrupt-entry, CIA dispatch and timer overhead.

  ## Results

  Measured in Copperline using an A500 configuration, a CIA-B level-6 interrupt every 10 ms, and the same scheduler-latency test disk.

| Measurement | Current AROS ROM | This PR | Kickstart 1.3 |
| --- | --- | --- | --- |
| Worst interrupt latency | 1,321 µs | 338 µs | 270 µs |
| Task switches/second | 776 | 2,623 | 3,039 |
| Signal round trips/second | 330 | 1,256 | 1,508 |
| Highest clean serial rate | 4,800 baud | 19,200 baud | 19,200 baud |

  Compared with the current AROS ROM:

  - Worst interrupt latency is reduced by approximately 74%, from 1,321 µs to 338 µs.
  - Worst latency is approximately 3.9× lower.
  - Task-switch throughput improves by approximately 3.4×.
  - Signal round-trip throughput improves by approximately 3.8×.
  - The highest tested serial rate without late samples increases from 4,800 to 19,200 baud.


warpdesign:
I noticed the null warnings: don't merge anything yet: I'll fix them in this PR

Kalamatee:
> I noticed the null warnings: don't merge anything yet: I'll fix them in this PR

Thanks - One structural request: could the new implementations simply replace the originals, rather than living in new dispatch_fast.S / switch_fast.S / preparecontext_fast.c / kernel_cpu_fast.c files that are only ever #included from the now-empty original files? There's no actual build-time selection happening — the comments say "build-time selected backend", but nothing selects anything — so the extra layer only adds indirection and breaks git blame for these core files. The old implementations remain in git history if we need to compare, and if a selectable classic/fast backend is genuinely planned, the split (with a real configure switch) can be introduced when it exists. That would also shrink the mmakefile change to just the genuinely new wait and signal_fast entries.

Kalamatee (approved):

PR #882, opened 2026-07-29
PFS3 i386 support / endianness fixes

bitplane:
PFS3 and PFSDoctor little-endian support

* add shared routines for PFS3 metadata and uses them in handler and PFSDoctor. 
* fix endian assumptions in anode numbers, dir entries, extra fields, volume names, boot blocks and list entry flags.
* PFSDoctor still gets the unconverted blocks when metadata has been overwritten or damaged, so invalid block IDs won't stop it
* CUnit tests cover block conversion, malformed metadata, recovery-mode directory entries and list entry layout. Failed conversions leave the original buffer unchanged.

~~**Tested on i386-pc only at the moment, hence draft PR.**~~

^ works now

~~I'm currently building a minimal i386-pc and m68k image and testing those. I'll attach them and mark this ready when I'm done proving them out :)~~

### DIFF
--- a/developer/debug/test/filesys/pfs3/cunit-pfs3-endian.c
+++ b/developer/debug/test/filesys/pfs3/cunit-pfs3-endian.c
@@ -7,6 +7,8 @@
 #include "blocks.h"
 #include "endian.h"
 
+void TestListTypeLayout(void);
+
 #define BLOCK_BYTES_1K 1024
 #define BLOCK_BYTES_2K 2048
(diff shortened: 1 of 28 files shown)


Kalamatee (approved):

bitplane:
okay I think this is ready. I've proven it locally, root boots in i386 and m68k, and I can get files in and out. Rebuilds are taking an age so I've not published nice clean images proving it

last commit was a fix for a double-release during dismount.

Issue #883, opened 2026-07-29
P96CTS 0.11 fails several tests in PAL mode, then leaves a garbled Workbench screen


codewiz:
**Describe the bug**

I revently ported my P96 Conformance Test Suite (P96CTS) to AROS, using the CGX API instead of Picasso96API.library.

It runs, but fails several tests even in a PAL AGA mode.

**To Reproduce**
1. Install p96cts-0.11.lha
2. Run "p96cts PAL 640x512x8"

You may want to redirect the output to a file because you won't be able to read the screen afterwards.
Also check the screen captures in output/

**Screenshots**

After running the suite, the system is still running, but the Workbench screen is corrupted. Looks like a copperlist issue, as the planes are aligned incorrectly:

<img width="1670" height="1418" alt="Image" src="https://github.com/user-attachments/assets/538ddbce-3bf6-46f1-a5e3-f7f5815f92cd" />

**Architecture**
 - Amiga (Copperline built from head)
 
**CPU**
 - m68k

**Version**
378596058676d758a17c8f14afd4ec68147f1bf4

codewiz:
This also happens after leaving the ScreenMode prefs Test screen.

codewiz:
This no longer happens. I guess it was fixed by one of the recent graphics.library / amigavideo / P96 fixes.

PR #886, opened 2026-07-29
graphics: fix RectFill drawmodes, plane masks and the stencil minterm

codewiz:
Four rendering bugs in `graphics.library` and `hidd.gfx`, found by running a
P96 driver conformance suite against AROS and comparing pixel for pixel with
what Picasso96's own software rasterizer produces for the same primitives.

Each is independent, and each is one commit.

### `RectFill()` drawmodes

Without an `AreaPtrn` the fill pattern is all ones, so `INVERSVID` clears it:
`JAM1` must then write nothing at all, and `JAM2` writes BPen. AROS treated
`INVERSVID` as a pen swap in both cases. Separately, `GetDriverData()` tests
`JAM2` before `COMPLEMENT`, so `JAM2 | COMPLEMENT` collapsed to a plain `JAM2`
fill and lost the inversion. Four of the eight mode combinations were wrong:

| mode | expected | was |
|---|---|---|
| `JAM1\|INVERSVID` | untouched | BPen |
| `JAM1\|INVERSVID\|COMPLEMENT` | untouched | inverted |
| `JAM2\|COMPLEMENT` | inverted | APen |
| `JAM2\|INVERSVID\|COMPLEMENT` | inverted | BPen |

### `rp->Mask` never reached a driver

`GetDriverData()` pinned `GC_COLMASK` to `~0`, so the write mask never left
graphics.library. The GC field and the generic per-pixel handling for it
already existed; nothing populated it. It is now set from `rp->Mask` for
palette bitmaps, and honoured in the p96gfx and amigavideo fills.

Left as `~0` on truecolor, where a plane mask has no counterpart, and for the
usual `0xFF`, so the existing `== ~0` fast paths still hit.

`setmaxpen.c` computes `rp->Mask` from the maximum pen precisely so that
rendering can skip the planes above it, which until now had no effect.

### `BltBitMap()` ignored plane counts

A blit covers `min(src, dst)` planes and skips the ones `mask` deselects. The
planar path in `bltbitmap.c` does both; the HIDD path did neither. It wrote
every destination plane, so a depth-3 source onto a depth-8 screen cleared the
top five planes instead of leaving them, and the `mask` argument was never
passed to `int_bltbitmap()` at all.

Both rules are now expressed as `GC_COLMASK`, which makes the two paths agree.

### `CopyBoxMasked()` did not implement the ANBC minterm

It is one of only the two minterms `BltMaskBitMapRastPort()` is defined for,
and the case was a stub that logged and fell through, so an inverted stencil
blit drew nothing.

What gets complemented is the pen, before the colormap lookup, not the color
the pen maps to. On a truecolor destination the two differ: a source pen of
`0x47` complements to `0xB8`, which maps to `00FCAA`, whereas inverting the
mapped color gives `FF0355`. The complement is masked to the source bitmap's
depth, since `~pen` on a `HIDDT_Pixel` would otherwise index the colormap far
out of bounds.

## Testing

Tested on m68k under an emulated A4000 with a Z3660 RTG board, at 640x480x8
and 640x480x24, against reference images captured from `rtg.library`.

Flipped from failing to passing at 8 bits: `RectFill-drawmodes`,
`BltBitMap-planemask`, `BltBitMap-shallow`, `BltBitMap-stencil` and
`BltBitMap-minterms`. Nothing regressed at either depth.

`BltBitMap-minterms` is worth a note: it fails on this board under AmigaOS
because the card driver's blit hooks ignore the minterm, and it passes here
because AROS falls through to the per-pixel path, which gets it right.

Still failing, and untouched by this PR: `BltTemplate-masks` and
`BltPattern-drawmodes` sweep `rp->Mask` through `PutTemplate` and
`PutPattern`, which need the same treatment as the fills; `BltBitMap-planemask`
and `-shallow` still fail at 24 bits, where the source pen has to be masked
before the ColorIndexMapping rather than the destination protected; and
`DrawLine-complement` is a bug in the card driver, failing identically under
AmigaOS.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/arch/m68k-amiga/hidd/amigavideo/amigavideo_bitmapclass.c
+++ b/arch/m68k-amiga/hidd/amigavideo/amigavideo_bitmapclass.c
@@ -559,7 +559,7 @@ VOID AmigaVideoBM__Hidd_BitMap__DrawLine(OOP_Class *cl, OOP_Object *o,
             }
         }
 
-        if (blit_fillrect(csd, data->pbm, x1, y1, x2, y2, fg, mode))
+        if (blit_fillrect(csd, data->pbm, x1, y1, x2, y2, fg, mode, GC_COLMASK(gc)))
             doSuper = FALSE;
     }
     if (doSuper)
@@ -989,7 +989,7 @@ VOID AmigaVideoBM__Hidd_BitMap__FillRect(OOP_Class *cl, OOP_Object *o, struct pH
          (mode == vHidd_GC_DrawMode_Copy)) ||
         ((height == 1) && (width >= 256) &&
          (mode == vHidd_GC_DrawMode_Copy)) ||
-        !blit_fillrect(csd, data->pbm, msg->minX, msg->minY, msg->maxX, msg->maxY, fg, mode)) {
+        !blit_fillrect(csd, data->pbm, msg->minX, msg->minY, msg->maxX, msg->maxY, fg, mode, GC_COLMASK(msg->gc))) {
         CMDDEBUGUNIMP(bug("[AmigaVideo:Bitmap] %s()\n", __func__);)
         OOP_DoSuperMethod(cl, o, (OOP_Msg)msg);
     }
--- a/arch/m68k-amiga/hidd/amigavideo/amigavideo_blitter.c
+++ b/arch/m68k-amiga/hidd/amigavideo/amigavideo_blitter.c
@@ -117,7 +117,7 @@ BOOL blit_copybox(struct amigavideo_staticdata *data, struct BitMap *srcbm, stru
     srcx2 = srcx + w - 1;
     dstx2 = dstx + w - 1;
     if (copy_minterm[mode] == 0xff)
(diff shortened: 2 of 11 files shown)


Kalamatee (approved):

warpdesign:
Did you do any tests on native display?

Kalamatee:
> Did you do any tests on native display?

I have done some minimal testing under WinUAE (native + p96), and on x86_64 (linux hosted and native), just to make sure that nothing was broken.

The only thing I have noticed is under WinUAE when it switches mode to the p96 mode, the cursor doesn't look like it is remapped to the hardware until it changes so displays incorrectly.

codewiz:
> Did you do any tests on native display?

P96CTS breaks things badly on native modes (issue #883). I'll get to it eventually.

> The only thing I have noticed is under WinUAE when it switches mode to the p96 mode, the cursor doesn't look like it is remapped to the hardware until it changes so displays incorrectly.

I see this with Z3660 too. After IPrefs runs, sometimes the curssor is corrupted, sometimes it's invisible.

I haven't filed a bug for this yet.

PR #891, opened 2026-07-29
detect FAT filesystems in GPT Basic Data partitions

bitplane:
"Microsoft Basic Data" partitions are a catch all that doesn't identify a filesystem. The GPT reader currently leaves these partitions without a DOS type, so FAT filesystems inside them don't get mounted.

This adds a shared FAT BPB parser and uses it to detect FAT12, FAT16 and FAT32 inside them. Existing explicit GPT mappings still take precedence; failed reads or invalid BPBs leave the DOS type unset.

The parser also replaces the separate BPB checks in the FAT handler, MBR reader and USB mass-storage class. The lightweight MBR-versus-superfloppy check stays permissive, but callers that need a filesystem type use full validation. So hopefully the only thing this will change is that GPT disks with those ambiguous partitions will load.

Tests run:

- [x] the attached unit tests
- [x] qemount’s generated FAT12, FAT16 and FAT32 filesystems
- [x] GPT Basic Data containing a non-FAT filesystem: partition discovered, no false mount
- [x] GPT EFI System Partition containing FAT32: explicit mapping still takes precedence.
- [x] Raw FAT12/16/32 “superfloppy” images: are are not mistaken for MBRs.
- [x] MBR with FAT16 and FAT32 partitions: partition table and volumes still mount normally.
- [x] MBR containing a non-FAT partition: MBR remains recognised without false mount.
- [x] Invalid BPB containing text such as FAT32, MSDOS or MSWIN: don't mount just because of labels
- [x] USB mass-storage raw FAT superfloppy
- [x] USB mass-storage MBR-partitioned FAT disk.
- [x] write + remount: data persists

### DIFF
--- a/compiler/fatbpb/bpb.c
+++ b/compiler/fatbpb/bpb.c
@@ -1,5 +1,5 @@
 /*
- * fat-handler - FAT12/16/32 filesystem handler
+ * FAT BIOS Parameter Block parsing link library
  *
  * Copyright (C) 2026 The AROS Development Team
  *
@@ -10,7 +10,7 @@
 #include <dos/filesystemids.h>
 #include <exec/types.h>
 
-#include "bpb.h"
+#include <linklibs/fatbpb.h>
 
 #define BPB_BYTES_PER_SECTOR       11
 #define BPB_SECTORS_PER_CLUSTER    13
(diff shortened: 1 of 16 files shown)


Kalamatee:
please move the BPB code into a linklib, and link the 3 components against that rather than including the files from another modules source directory.

Kalamatee (approved):

PR #894, opened 2026-07-29
graphics:  Fix RectFill INVERSVID regression affecting scrolling

### LINKED ISSUE #893: Regression: scrolling broken in DOpus 4 / A500 configuration
**Describe the bug**
When reading a text file in DOpus 4 in A500 configuration, scrolling is broken: graphics are trashed when scrolling.

**To Reproduce**
1. Open DOpus 4
2. Double-click on a text file top open the built in viewer
3. Scroll using the cursor keys

**Expected behaviour**
Scrolling should work without producing graphic artifacts

**Screenshots**
Before commit a9072ed70386db03e7247c725a5f54b60483c11c:

https://github.com/user-attachments/assets/79c6529f-bcd3-4cf5-84bb-b0f60a05a712

After:

https://github.com/user-attachments/assets/b6531a0d-1b2e-4ede-a92b-8f24ffe384a5

**Architecture**
 - Amiga m68k, A500 configuration with 1Mb fast ram
 
**CPU**
 - m68k

**Version**
a9072ed70386db03e7247c725a5f54b60483c11c

**Additional context**
It broke in this exact commit, it works in the previous commit.

DOpus 68k is open source and can be freely distributed here: http://dopus.free.fr/download.php

warpdesign:
## Summary

  Fixes a graphics regression introduced by `a9072ed70386db03e7247c725a5f54b60483c11c`.

  That change made `RectFill()` with `JAM1 | INVERSVID` and no `AreaPtrn`
  perform no drawing. However, existing graphics.library functions including
  `ScrollRaster()`, `ClearEOL()` and `ClearScreen()` deliberately toggle
  `INVERSVID` before calling `RectFill()` to fill an area with the background
  pen.

  As a result, DOpus 4 scrolling no longer cleared newly exposed rows. New text
  was drawn over stale pixels, producing severe visual corruption.

  ## Changes

  - Restore the established background-pen fill behavior for `INVERSVID`
    without an `AreaPtrn`.
  - Retain the `COMPLEMENT` support added by the original change.
  - Save and restore the shared HIDD GC draw mode after temporarily selecting
    `Invert`, preventing subsequent operations from inheriting it.

  ## Testing

  Tested with DOpus 4 using an A500 AROS ROM configuration. Scrolling now clears
  and redraws exposed rows correctly without the previous corruption.

  The author of the original change is mentioned for review because this
  restores the previous `JAM1 | INVERSVID` semantics relied upon by existing
  graphics.library callers.

After the fix:

https://github.com/user-attachments/assets/9c7de56c-a8e8-42e0-982a-8600b0cd2b5f

@codewiz I'll wait for your confirmation that I did not break anything you fixed in the original PR before removing the draft state :)

Fixes #893

### DIFF
--- a/rom/graphics/rectfill.c
+++ b/rom/graphics/rectfill.c
@@ -68,38 +68,31 @@
             /* When rastport has areaptrn, let BltPattern do the job */
             BltPattern(rp, NULL, xMin, yMin, xMax, yMax, 0);
         } else {
-            /*
-             * Without an AreaPtrn the fill pattern is all ones, so INVERSVID
-             * clears it: JAM1 then writes nothing at all, and JAM2 writes
-             * BPen everywhere. COMPLEMENT inverts the destination wherever
-             * the mode writes, whichever pen it would have used.
-             */
+            /* Without an AreaPtrn, INVERSVID selects the background pen. */
             BOOL inversvid = (rp->DrawMode & INVERSVID) != 0;
-            BOOL jam2 = (rp->DrawMode & JAM2) != 0;
+            OOP_Object *gc  = GetDriverData(rp, GfxBase);
+            struct Rectangle rr;
+            HIDDT_Pixel oldfg = GC_FG(gc);
+            HIDDT_DrawMode olddrmd = GC_DRMD(gc);
 
-            if(jam2 || !inversvid) {
-                OOP_Object *gc  = GetDriverData(rp, GfxBase);
-                struct Rectangle rr;
-                HIDDT_Pixel oldfg = GC_FG(gc);
-
-                if(inversvid) {
-                    GC_FG(gc) = GC_BG(gc);
-                }
-                if(rp->DrawMode & COMPLEMENT) {
-                    GC_DRMD(gc) = vHidd_GC_DrawMode_Invert;
-                }
-
-                /* This is the same as fillrect_pendrmd() */
-
-                rr.MinX = xMin;
-                rr.MinY = yMin;
-                rr.MaxX = xMax;
-                rr.MaxY = yMax;
-
-                do_render_with_gc(rp, NULL, &rr, fillrect_render, NULL, gc, TRUE, FALSE, GfxBase);
-
-                GC_FG(gc) = oldfg;
+            if(inversvid) {
+                GC_FG(gc) = GC_BG(gc);
             }
+            if(rp->DrawMode & COMPLEMENT) {
+                GC_DRMD(gc) = vHidd_GC_DrawMode_Invert;
+            }
+
+            /* This is the same as fillrect_pendrmd() */
+
+            rr.MinX = xMin;
+            rr.MinY = yMin;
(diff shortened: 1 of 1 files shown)


codewiz:
Sorry for the regression, your diagnosis looks right to me.

I should get to a point where I can run P96CTS on native modes too, so anything touching graphics.library would be covered.

codewiz:
Ah, but P96CTS lacks coverage for ScrollRaster at this time, so it wouldn't have caught this anyway...

I'll add it as soon as I'm done fixing the current test failures on Z3660 on both AROS and AOS.

codewiz:
I asked Claude to review this, and it agrees that this fixes the regression with DOpus4, but also warned me that it will re-introduce the RectFill-drawmodes failure that a9072ed70386db03e7247c725a5f54b60483c11c was fixing.

I will let Claude speak for itself:

----

Where I'd differ is the location. The patch also restores the two cases the original change fixed, so JAM1|INVERSVID fills BPen and JAM1|INVERSVID|COMPLEMENT inverts, where AmigaOS leaves the area untouched. The no-op isn't incidental: with no AreaPtrn the fill pattern is all ones, INVERSVID clears it, and JAM1 writes only where the pattern is set — so there is nothing to write. That's what P96's rasteriser produces, and it's what RectFill-drawmodes in my conformance suite checks.

The assumption that needs fixing lives in the callers, and text.c:575 already shows the right shape — it guards the same XOR idiom with if ((rp->DrawMode & ~INVERSVID) == JAM2), so it only ever reaches a mode that fills. The other four sites are missing that guard. My suggestion is to force JAM2 there and let INVERSVID keep choosing the pen:

/* RectFill() draws nothing in JAM1 with the pattern inverted, so ask for
   JAM2 and let INVERSVID pick the pen. */
SetDrMd(rp, JAM2 | ((oldDrMd ^ INVERSVID) & (INVERSVID | COMPLEMENT)));

Working through every input mode, that reproduces the pre-regression result at each one — JAM1->BPen, JAM1|INVERSVID->APen, JAM2->BPen, JAM2|INVERSVID->APen, COMPLEMENT inverting — while leaving RectFill() conformant. The single deliberate difference is JAM2|COMPLEMENT, which used to fill BPen because GetDriverData() tests JAM2 before COMPLEMENT and now inverts correctly.

----

I'd feel more comfortable if we simply matched AmigaOS, but first I'd have to write a testcase...

codewiz:
I'm proposing this alternative fix:
https://github.com/aros-development-team/AROS/pull/896

It's also a draft because I don't have P96CTS coverage and I didn't even run DOpus4 to see if it fixes your regression.

warpdesign:
Thanks! I'll try later if it fixes the DOpus regression.

--
Nicolas RAMZ - http://www.warpdesign.fr/

AthenaJS - https://athenajs.github.io

Le jeu. 30 juil. 2026, 05:19, Bernie Innocenti ***@***.***> a
écrit :

Issue #900, opened 2026-07-30
ScreenMode Prefs window does not fit vertically in a PAL High-Res (640x256) screen


codewiz:
**Describe the bug**

This makes the buttons at the bottom inaccessible. Moreover, using the menu items does not seem to work.

**Screenshots**
<img width="2277" height="1822" alt="Image" src="https://github.com/user-attachments/assets/32ec16e9-b0ea-41ed-ae7b-7507e6bf5ff6" />

**Architecture**
 - Amiga
 
**CPU**
 - m68k

codewiz:
I made it a lot smaller, but it's still too tall for an NTSC 640x200 screen:

<img width="692" height="540" alt="Image" src="https://github.com/user-attachments/assets/14a8b1c1-f733-40e6-93d1-8889e5a49e73" />

kowoba:
This could largely be fixed by using less vertical space on widgets, gadgets and layout - why do AROS Intuitions insist on so much space above and below titles of windows, buttons, groups etc? On RTG with 1:1 yes, but on 1:2-ish PAL/NTSC it looks awkward and takes precious space.

codewiz:
> This could largely be fixed by using less vertical space on widgets, gadgets and layout - why do AROS Intuitions insist on so much space above and below titles of windows, buttons, groups etc? On RTG with 1:1 yes, but on 1:2-ish PAL/NTSC it looks awkward and takes precious space.

I agree. Is there a way to make MUI / Zune adjust vertical spacing based on the aspect ratio of the current mode?

codewiz:
Anyway, I got the window to fit within 640x200 by dropping the unnecessary labels at the top and inlining Autoscroll with Depth.

PR #901, opened 2026-07-31
prefs/screenmode: fit the editor on a 200-line screen

codewiz:
The editor's minimum height was larger than a 640x256 PAL screen, so the
Save/Use/Cancel buttons could not be reached (#900). It now fits 640x200
NTSC with room to spare.

The height went into parts of the layout that could not shrink:

- A `Scrollgroup` around the mode features, whose minimum is a vertical
  scrollbar plus a horizontal one, for a handful of lines of text.
- Three all-`RectangleObject` spacer rows in the attributes table.
- `Label1()`/`LLabel1()` on every cell of that table. Those reserve an
  ImageButton frame, meant for labelling a framed gadget; the cells are
  plain text, so they use the frameless variants now.
- The "Display Mode" and "Mode Attributes" headings, which repeat what
  the two panels obviously are.
- Depth and Autoscroll on separate rows, each wasting the two columns
  that hold the checkmark and its label in the rows above.

`MSG_DISPLAY_MODE` and `MSG_MODE_ATTRIBUTES` stay in the catalog even
though nothing uses them now, since removing entries renumbers it and
invalidates the translations.

The decorative orientation image is untouched by this patch but did not
load in the test environment, which is why it is absent from the
screenshot. A brush that fails to load collapses to zero height, since
`__area_finish_minmax()` clamps `MUIA_FixHeight` against a zero min and
max. Being looked at separately.

Tested on m68k under Copperline at 640x256 PAL and 640x200 NTSC.

### DIFF
--- a/workbench/prefs/screenmode/smattributes.c
+++ b/workbench/prefs/screenmode/smattributes.c
@@ -28,7 +28,7 @@ struct SMFloattext_DATA
 
 IPTR SMFloattext__MUIM_DrawBackground(Class *CLASS, Object *self, struct MUIP_DrawBackground *message)
 {
-    DoMethod(_parent(self), MUIM_DrawBackground, message->left, message->top, message->width, message->height, message->xoffset, message->yoffset, message->flags);
+    return DoMethod(_parent(self), MUIM_DrawBackground, message->left, message->top, message->width, message->height, message->xoffset, message->yoffset, message->flags);
 }
 
 ZUNE_CUSTOMCLASS_1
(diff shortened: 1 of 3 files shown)


codewiz:
Looks like this now:

aros-screenmode

codewiz:
It barely fits a 640x200 screen after running the full Startup-Sequence and giving IPrefs a chance to set the fonts:

image

Kalamatee (approved):

PR #904, opened 2026-07-31
pcixhci: build without debug output by default

jonx:
The driver is currently built with `-DDEBUG=1 -DPCIUSB_XHCI_DEBUG`, so every system carrying `pcixhci.device` prints the full xHCI transfer trace at runtime and pays for the strings in the binary.

This was my mistake: the flag was flipped in fb10baf2e9, a commit that had nothing else to do with the driver and should never have touched this file. Sorry about that.

Restoring the commented-out form takes the module from 190,648 bytes back to 131,960 on aarch64, and stops the trace.


jonx:
Companion: #905 (split-transaction hub addressing), same driver, independent change.

PR #906, opened 2026-07-31
graphics: delete the old MoveRaster() blitting loop

codewiz:
Disabled behind `#if USE_OLD_MoveRaster` since it was replaced, and it no longer compiles. No functional change.


Kalamatee (approved):

codewiz:
@Kalamatee as discussed.

PR #914, opened 2026-08-02
con-handler: run over a device, so AUX: works

codewiz:
Completes the AUX: handler added in #911, which until now had nothing to hand its startup packet to. `AUX0:` gives a shell over a serial line: reads return on a line instead of waiting for the buffer to fill, and Ctrl-C reaches the running command, neither of which `SER:` can do.

No console code was duplicated. con-handler already drives anything answering `CMD_READ`/`CMD_WRITE` -- the packet loop names console.device nowhere -- so only its startup needed to say which device. Line editing, `ACTION_WAIT_CHAR` and ^C..^F come along unchanged.

**genmodule** -- `Resident=` did not do what its name says. It enqueued a FileSysEntry, and the `AddSegment()` that turns those into named segments runs once, while dos.library initialises. Any handler coming up after dos was therefore never findable by name, which is every handler needing dos calls to build its seglist. It now registers the segment directly, opening dos.library rather than assuming it, so a module initialising earlier still falls back to the existing scan. This also fixes port-handler, which RunHandler had to `LoadSeg` because the lookup always missed.

**Two pre-existing bugs in con-handler's open-failure path**, both memory corruption, neither previously reachable because opening a window does not really fail:

- the read IORequest was freed, and then `close_con()` freed it again
- the filehandle `close_con()` had just freed was returned to the caller, which saw success and ran the entire handler on freed memory

The second one is why the failures looked so strange while debugging: the startup packet was answered `DOSTRUE` from the failure path, so the handler reported itself online and then faulted somewhere else entirely.

**Device mode** additionally needed: a read request sized for the device rather than for console.device (serial.device refuses anything shorter than an `IOExtSer`), a read of one character rather than a bufferful since a plain device will not answer until the whole count has arrived, and three places that assumed a window -- RAW: selected from a `dn_Startup` that is now a pointer, intuition.library and input.device required outright, and the packet loop opening a window on demand.

The mountlist is `AUX0` rather than `AUX` because a file of that name cannot be checked out on Windows. Numbering matches SER0 and leaves room for further units.

Tested on m68k over the Paula serial driver: input, output and Ctrl-C all work, with the mountlist naming the device explicitly.

Known gaps, both for later: `Baud` and `Control` reach the handler and are not yet applied (no `SDCMD_SETPARAMS`), and reads are one character at a time -- `SDCMD_QUERY` reports how many more are buffered, so a burst can be drained in one further read.


codewiz:
Pushed `con-handler: shut down on ACTION_DIE`.

A device-backed console is one shared stream, so it now publishes its port in `dn_Task` and handles `ACTION_INHIBIT`/`ACTION_DIE`. CON: still leaves `dn_Task` clear, which is what gets each open its own window.

Two things this uncovered:

- `ACTION_DISK_INFO` reported `fh->conreadio` as `id_InUse`. That is non-NULL for as long as the handler lives, so the console always looked busy and could never be dismounted. It now reports the number of open streams.
- A startup failure was replied as plain DOSFALSE with no error code, so `RunHandler` could only say `handler failed startup [0]`.

Pairs with #916, which stops Assign from being the thing that sends `ACTION_DIE` in the first place.

Kalamatee (approved):

PR #923, opened 2026-08-03
Raspi boot console and config files

bsek:
- Fixes console output during boot on aarch64 
- Adds dtb files for pi4 and zero 2w
- Removes overlay references in config.txt which are not used


Kalamatee (approved):

Kalamatee:
You may want to make it also grab the Pi5 files

PR #925, opened 2026-08-03
Fix/amiga classic live boot

bitplane:
A few fixes needed to get the floppy/ISO combo booting on an A500 with 2MB chip / 8MB fast under UAE:

1. fix bootstrap path
2. deadlock caused by wb reset
3. fonts eating all the RAM


bitplane:
Oops found some bugs. Moved to draft for a moment

Kalamatee:
regarding the iprefs changes - I suspect you should be using process->pr_WindowPtr to signal/check if the request should be displayed (the normal OS way of handling this), rather than adding initial_load.

Kalamatee (approved):

PR #1003, opened 2026-08-17
NListtree.mcc: fix heap corruption in tree image string building

retrofaza:
strncat() with n set to the full size of the destination buffer can write past its end when data->buf is almost full, overwriting the header of the adjacent heap block. With enough memory (e.g. hosted AROS with -m 4096) this caused crashes in tlsf_freevec() when the corrupted blocks were freed, and was reproducible in Prefs/Zune -> NListtree -> Expand.

Restore the upstream NList behaviour (commit 8d073d73): format each image spec into a small tmp[64] buffer and append it with the bounded strlcat(). This matches the upstream code so the files stay in sync and are easier to update.

### DIFF
--- a/workbench/classes/zune/nlist/nlisttree_mcc/NListtree.c
+++ b/workbench/classes/zune/nlist/nlisttree_mcc/NListtree.c
@@ -2795,15 +2795,15 @@ static void InsertTreeImages( struct NListtree_Data *data, struct MUI_NListtree_
 
     if ( ( x1 != -1 ) && x2 )
     {
-      char tmpbuf[DATA_BUF_SIZE];
+      char tmp[128];
 
       if ( x1 == SPEC_Space )
         otn->tn_Space += x2 + data->IndentWidth;
 
       if ( otn->tn_Space > 0 )
       {
-        snprintf(tmpbuf, DATA_BUF_SIZE, "\033O[%p;%x;%d,%d]", (APTR)data->Image[IMAGE_Tree].ListImage, (unsigned int)MUIA_TI_Spec, (int)SPEC_Space, (int)otn->tn_Space );
-        strncat(data->buf, tmpbuf, DATA_BUF_SIZE);
+        snprintf(tmp, sizeof(tmp), "\033O[%p;%x;%d,%d]", (APTR)data->Image[IMAGE_Tree].ListImage, (unsigned int)MUIA_TI_Spec, (int)SPEC_Space, (int)otn->tn_Space);
+        strlcat(data->buf, tmp, DATA_BUF_SIZE);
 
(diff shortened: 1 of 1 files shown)


Kalamatee:
```
#  char tmp[64] is sized by eyeball and can truncate on 64-bit. The longest format in the patch, "\033O[%lx]\033O[%lx;%x;%d,%d]", needs 74 bytes worst case on 64-bit (58 on 32-bit) - I measured it. Realistic values fit comfortably (~53 bytes: pointers are ≤12 hex digits, the two ints are a small SPEC_* and a literal 3), so this isn't a live bug, but there's no headroom by construction and snprintf will silently truncate if any of those values ever grows. 96 or 128 bytes would cost nothing.
 
# Truncation is now silent. strlcat's return is discarded, so once the 4096-byte buffer fills, escapes are dropped - possibly cut mid-sequence, leaving a partial \033O[… with no closing ]. I checked the parser: it scans to '\0' and simply doesn't render, so there's no crash or over-read — but the row renders wrong with no diagnostic. A D() warning when strlcat returns ≥ DATA_BUF_SIZE would make the next report of this much easier.
 
# (unsigned long) for a pointer is correct on every AROS ABI but is a step down in self-documentation from %p/APTR. IPTR with %lx, or keeping %p, would express intent better. Nil practical risk today (it only breaks on an LLP64 ABI, which AROS doesn't target).
 
# Signedness inconsistency: %d is fed (unsigned int)SPEC_Hor, (unsigned int)SPEC_Space and (unsigned int)data->IndentWidth, while neighbouring arguments use (int). Harmless (no warning without -Wformat-signedness), but arbitrary.
 
# Readability regressions in the refactor: dropping the ln local in favour of inlining (struct List *)&(CLN(otn))->ln_List into an already-long condition, and folding the Open/Closed choice into the array subscript. Both are equivalent — I checked the logic carefully - just harder to read.
```

IMHO it would have been better to push the corrected code upstream than to regress AROS's code to match upstream.

The %p form this replaces wasn't an AROS quirk - it was a deliberate correction (ac5da7ae3f, "s/%lx/%p for pointers"). %p with an APTR says "this is a pointer" and stays correct whatever the ABI does; %lx with (unsigned long) only works as long as sizeof(unsigned long) == sizeof(void *) holds, which is an assumption we'd rather not re-acquire. The buffer overflow fix is the valuable part of this patch and stands entirely on its own - strncat()'s third argument bounds the source, not the destination, so it never protected data->buf at all - and none of the consumers care which of the two formats is used: NList parses the value with strtoul(..., 16) and stpncpy_noesc() just skips to the ].

Kalamatee (approved):

PR #1006, opened 2026-08-18
m68kemu: Initial experimental implementation of m68kemu.library for transparent m68k binary emulation

metaneutrons:
## Overview

This PR proposes an initial/experimental implementation of **`m68kemu.library`**, an emulation layer allowing classic AmigaOS m68k hunk binaries to run transparently on non-m68k AROS architectures (x86_64, AArch64, etc.).

The goal is to provide a baseline for running standard OS-friendly m68k CLI tools and utilities directly from Shell/Wanderer without requiring a full external emulator like UAE.

## How It Works

* **CPU Core:** Uses [Moira](https://github.com/dirkwhoffmann/Moira) (Dirk W. Hoffmann, MIT license) as a git submodule (`rom/m68kemu/Moira`), configured in M68040 mode.
* **Containment & Memory:** Executes within a contained 32 MB big-endian sandbox with its own first-fit heap allocator for `AllocMem`/`FreeMem`.
* **API Thunking & Struct Translation:** Intercepts `JSR -offset(A6)` library jumps via A-line trap opcodes. Provides native thunks for ~910 common API functions across 17 core libraries (`exec`, `dos`, `intuition`, `graphics`, `utility`, etc.) with automated 32-bit $\leftrightarrow$ 64-bit shadow struct layout translation.
* **Transparent `dos.library` Hook:** `RunCommand` and `CreateNewProc` check `GetSegListInfo(GSLI_68KHUNK)`. If an m68k hunk is detected and `m68kemu.library` is available, execution is routed through `RunHunk()`.
* **Safety / Fallback:** If `m68kemu.library` is not available or if the binary is native ELF, `dos.library` behaves exactly as before. Native m68k builds automatically skip compiling this subsystem (`ifneq ($(AROS_TARGET_CPU),m68k)`).

## Current Status & Tested Software

The implementation is verified via both host-side unit tests and a selection of classic software:

* **Host Test Suite (`tests/m68kemu_host/`):**
  * `test_m68kemu`: Context lifecycle, memory sandbox, big-endian heap allocator, pointer translation (6/6 tests pass).
  * `test_moira_exec`: Basic M68040 instruction execution, loops, memory access (4/4 tests pass).
  * `test_hello_thunks`: End-to-end integration (`OpenLibrary` $\rightarrow$ `PutStr` thunk $\rightarrow$ clean exit).
* **Initial Real-World Software Tested:**
  * CLI tools: LZX, LhA, DMS, SysInfo, Which.
  * GUI applications: Directory Opus 4.12, AmigaTerm 1.0, Textcraft Plus.
  * Dynamic loading: Third-party m68k shared libraries from `LIBS:`.

## Known Limitations & Need for Real-World Testing

Amiga software is notoriously diverse, and this implementation is intentionally focused on **OS-compliant, system-friendly binaries**. It is not intended as a replacement for full custom-chip emulators (e.g. UAE).

Current limitations include:
* **Custom Chips:** Only basic raster beam reading (`VPOSR`/`VHPOSR`) and `POTGOR` are stubbed. There is no planar blitter, copper, audio DMA, or sprite hardware emulation.
* **Concurrency:** Emulated CPU execution is currently single-threaded.
* **Memory Limits:** Fixed 32 MB containment sandbox per context; `MEMF_CHIP` vs `MEMF_FAST` attributes are currently treated uniformly within containment.
* **Thunk Coverage:** While ~910 functions across 17 libraries are mapped, less common LVOs and esoteric struct tags will need ongoing refinement.

**Wider community testing, bug reports, and feedback on architectural direction or edge cases are very welcome!**


rom/dos/createnewproc.c
Kalamatee, line 1: this should probably be guarded by -:
```

#if !defined(__mc68000__)
 
#endif

```
It doesn't make much sense routing it on m68k targets.

rom/dos/runcommand.c
Kalamatee, line 1: same as createnewproc - it should probably be guarded so its only used on non-m68k targets.

rom/dos/createnewproc.c
metaneutrons: Yikes! Yes, you're right. Fixed.

Kalamatee (approved):

rom/dos/runcommand.c
metaneutrons: Yikes! Yes, you're right. Fixed.

PR #1013, opened 2026-08-19
hidds/v3d: Add VideoCore VI (V3D 4.2) GPU driver with Mesa OpenGL

metaneutrons:
## Summary
Adds an experimental native Gallium GPU driver (`hidd.gallium.v3d`) for the Broadcom VideoCore VI (V3D 4.2) GPU on BCM2711 (Raspberry Pi 4), providing hardware-accelerated OpenGL 3.3 and OpenGL ES 3.1 via Mesa.

## Architecture
- **Hardware Layer (`v3d_hw.c`)**: Direct MMIO register access (`0xFEC00000` / `0xFEC04000`), Buffer Object management in contiguous memory (`MEMF_31BIT`), and Control List Executor (CLE) job submission for binning and rendering.
- **DRM Ioctl Shim (`v3d_drm_shim.c`)**: User-space DRM interface shim bridging Mesa V3D driver calls directly to hardware without Linux kernel DRM.
- **Gallium HIDD (`v3d_galliumclass.c`)**: Integrates with AROS Gallium state tracker to instantiate `v3d_screen_create()`.

### DIFF
--- a/workbench/hidds/v3d/drm-stubs/drm-uapi/drm_fourcc.h
+++ b/workbench/hidds/v3d/drm-stubs/drm-uapi/drm_fourcc.h
@@ -1,0 +1,15 @@
+/* DRM fourcc stub for AROS */
+#ifndef DRM_FOURCC_H
+#define DRM_FOURCC_H
+#include <stdint.h>
+#define fourcc_code(a, b, c, d) ((uint32_t)(a) | ((uint32_t)(b) << 8) | ((uint32_t)(c) << 16) | ((uint32_t)(d) << 24))
+#define DRM_FORMAT_MOD_INVALID 0x00ffffffffffffffULL
+#define DRM_FORMAT_MOD_LINEAR 0
+#define DRM_FORMAT_MOD_BROADCOM_UIF (((uint64_t)0x09) << 56 | 2)
+#define DRM_FORMAT_MOD_BROADCOM_SAND128 (((uint64_t)0x09) << 56 | 3)
+#define DRM_FORMAT_ARGB8888 fourcc_code('A','R','2','4')
+#define DRM_FORMAT_XRGB8888 fourcc_code('X','R','2','4')
+#define DRM_FORMAT_ABGR8888 fourcc_code('A','B','2','4')
+#define DRM_FORMAT_XBGR8888 fourcc_code('X','B','2','4')
+#define DRM_FORMAT_RGB565 fourcc_code('R','G','1','6')
+#endif
--- a/workbench/hidds/v3d/drm-stubs/drm-uapi/v3d_drm.h
+++ b/workbench/hidds/v3d/drm-stubs/drm-uapi/v3d_drm.h
@@ -1,0 +1,275 @@
+/*
+ * Copyright © 2014-2018 Broadcom
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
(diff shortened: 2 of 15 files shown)


bsek:
Hi, you should look at how the VC4 driver was made.

**The problem: two copies of the NIR compiler in one process**

`hidd.gallium.v3d`'s `mmakefile.src` links Mesa's V3D gallium driver and the Broadcom NIR-to-VIR compiler (`nir_to_vir`, `vir`, `vir_to_qpu`, `qpu_schedule`, ...) statically into the hidd module itself (`alwayscxxlink=yes uselibs=""`). If `mesa3dgl.library` is loaded in the same process — which it will be, since it's the Mesa state tracker that creates GL contexts — you now have **two independently-linked copies** of the NIR compiler stack running side by side.

That's fine for some drivers and fatal for others, depending on what crosses the module boundary.

**Why TGSI drivers get away with it and NIR drivers don't**

Classic Gallium drivers (softpipe, nouveau, vmware) receive shaders from the state tracker as flat TGSI token arrays — plain data, no ownership semantics. Each side allocates and frees its own copies; nothing shared crosses the boundary. A driver linking its own private copy of the compiler/aux code in that model is harmless, because that copy never touches the state tracker's objects.

NIR-based drivers (both vc4 and v3d/v3d42 use the same `src/broadcom/compiler` pipeline) work differently: the state tracker hands the driver a `nir_shader *` and **transfers ownership** of it. That object is:
- allocated with `ralloc`, a hierarchical allocator with its own bookkeeping
- full of `glsl_type *` pointers into Mesa's process-global type table (a set of static singletons, e.g. built-in scalar/vector/matrix types, registered once per module)

If the driver lives in a separate module from the state tracker, each module gets its *own* copy of both `ralloc`'s internal state and the `glsl_type` singleton table (because they're statically linked, not shared). The driver then:
1. calls `ralloc_free()` on an object that was allocated by the *other* module's allocator — a cross-module free, undefined behavior
2. compares `glsl_type` pointers against its own table's addresses, which never match the state tracker's table — this is exactly the class of bug NIR's internal validator is designed to catch

We hit this empirically on the vc4 driver: the first attempt used exactly this shape (driver in the hidd, state tracker in mesa3dgl.library, calling across via `CreatePipeScreen`/`DisplayResource`), and it crashed on this. Disabling the NIR validator didn't fix the underlying corruption, it just stopped it from being caught early.

You should adopt the same GalliumCoreAPI-style runtime binding vc4 uses, so the driver can ship as an independent hidd module without duplicating the NIR/ralloc/glsl_type state.

bsek:
And one small detail: the Author line you add to the source files is unnecessary. Git already knows who wrote the file, and if you add it to an existing file, it may be inaccurate.

metaneutrons:
Hi @bsek,

thank you very much for the insightful and detailed architectural review! 

You are completely spot-on regarding the dual NIR compiler instance and cross-module `ralloc` / `glsl_type` ownership issue between `hidd.gallium.v3d` and `mesa3dgl.library`.

I have updated the PR accordingly:
1. **GalliumCoreAPI Runtime Binding:** Refactored `mmakefile.src` to integrate with `galliumglue.py` and produce `libgallium_v3d_gca.a`. All Mesa core / NIR imports now route transparently through the `GalliumCoreAPI` trampoline table, and `v3d_galliumclass.c` binds it at runtime via `gca_bind()` from `aHidd_Gallium_CoreAPI`.
2. **Clean Headers:** Removed the redundant `Author:` tags across all source and header files to follow standard AROS copyright conventions.

The branch has been updated with these changes. Thanks again for pointing this out and guiding the design towards the proper GalliumCoreAPI pattern!

PR #1018, opened 2026-08-19
m68k-amiga: cd.device fixes to make CD32 discs bootable

LinuxJedi:
While working on CD32 support in the Copperline emulator I found that the AROS ROM can no longer boot CD32 discs, and traced it to a chain of problems in cd.device. This series fixes them.

The headline bug is a one-liner: 34d67c894690 switched the m68k ROM's CD filesystem from the cdfs handler to CDVDFS back in 2019, but cd.device still registers its boot node with DosType 'CDFS'. CDVDFS registers itself as 'CDVD', and CliInit only falls back to handler-name matching when the DosType is zero, so CD0: has been unmountable ever since.

Behind that were a few driver bugs in the Akiko command path:

* CD32_Cmd enables RXDMA/TXDMA before clearing the previous command's latched completion bits, so the stale latches fire the interrupt handler immediately, which disables the enables again and steals the completion signal. The retry loop also never re-enables RXDMA after consuming an unsolicited response (media change, SUBQ), which can park the unit task in an untimed Wait() forever.
* The initial media/TOC probe runs from InitCode(RTF_COLDSTART) at residentpri 5, so any drive misbehaviour wedges the boot task on the boot splash with no diagnostic. The series moves it onto the unit task via a new optional uo_Init hook.
* Neither drive wait had a timeout.
* The published DosEnvec carried stack garbage in de_Mask and de_Baud (the copy was off by one, de_TableSize being the highest valid index rather than a count), and the hand-built timer reply port never had its message list initialised even though ReplyMsg() enqueues on it.

With the series applied a CD32 disc mounts through CDVDFS and boots - real CD32 discs have no boot block, just an ISO9660 volume with an S/Startup-Sequence, so nothing in dosboot needed to change. Verified under Copperline against Pinball Fantasies, which now loads from CD and runs. Where this driver and the CD32 Kickstart cd.device disagreed about the wire protocol I checked both against the Kickstart ROM's parsing (for example the media-status byte: KS masks it with 3, this driver compares the whole byte against 0x83 - the real drive sends 0x83, and the emulator now does too). This is all under emulation, so a test on real hardware would be very welcome.

Two things I noticed nearby but did not touch, in case anyone wants them: the m68k level-2 wrapper acks INTF_PORTS whether or not any server claimed it (Amiga_Level_6 similarly handles INTF_EXTER|INTF_INTEN but only ever acks INTF_EXTER), and once discs mount, Pinball Fantasies gets as far as opening its screen and then loops on AmigaVideoBM__Root__New failing to instantiate a 640x256 depth-2 bitmap, which looks like a separate amigavideo issue.

### DIFF
--- a/arch/m68k-amiga/devs/cd/cd.c
+++ b/arch/m68k-amiga/devs/cd/cd.c
@@ -37,6 +37,14 @@ static VOID cdTask(IPTR base, IPTR unit)
     struct IOStdReq *io;
 
     D(bug("%s.%d Task, Port %p\n", cu->cu_UnitOps->uo_Name, cu->cu_Unit, cu->cu_MsgPort));
+
+    /* Blocking drive probes belong here, not in resident init: a
+     * misbehaving drive must cost a failed mount, not a wedged boot
+     * task. I/O queued by early openers is served after this returns.
+     */
+    if (cu->cu_UnitOps->uo_Init)
+        cu->cu_UnitOps->uo_Init(cu->cu_Private);
+
     do {
         WaitPort(cu->cu_MsgPort);
         io = (struct IOStdReq *)GetMsg(cu->cu_MsgPort);
--- a/arch/m68k-amiga/devs/cd/cd.c
+++ b/arch/m68k-amiga/devs/cd/cd.c
@@ -164,7 +164,10 @@ static int cd_Init(LIBBASETYPE *cb)
     NEWLIST(&cb->cb_Units);
     InitSemaphore(&cb->cb_UnitsLock);
 
-    /* Hand-hacked port for timer responses */
+    /* Hand-hacked port for timer responses. ReplyMsg() enqueues on
+     * mp_MsgList before signalling, so it must be a valid empty list.
+     */
+    NEWLIST(&cb->cb_TimerPort.mp_MsgList);
     cb->cb_TimerPort.mp_SigBit = SIGB_SINGLE;
     cb->cb_TimerPort.mp_Flags = PA_SIGNAL;
     cb->cb_TimerPort.mp_SigTask = FindTask(NULL);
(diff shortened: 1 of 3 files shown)


Kalamatee (approved):

LinuxJedi:
Picked up the Copilot findings in d9f7a465: the interrupt server now reports itself as handler when it acted on TXDMA or PBX status (exec's m68k IntServer discards the result, but no reason not to keep the contract honest), and CD32_CmdRead disables the PBX interrupt on its way out so a late sector cannot signal the unit task under some later Wait().

The shared timer request I left alone: it is not actually racy, since every user - the command timeouts and cdDelayMS() in the TOC loop - runs on the one unit task, which owns all drive I/O. That invariant is now documented next to the request so whoever adds a second backend with its own unit task knows to give it a per-unit request.

Kalamatee:
Looks good to me. My only thoughts are regarding -:
image
..and if this could be detected at runtime (e.g. in the module init?) instead of using hard coded values?

Either way im happy to merge this

LinuxJedi:
>  ..and if this could be detected at runtime (e.g. in the module init?) instead of using hard coded values?

Looking into it now to see what I can do.

LinuxJedi:
Good idea - done in e3eb2b54. The DosType now comes from FileSystem.resource at runtime, preferring 'CDVD' and falling back to the old cdfs handler's 'CDFS', so the boot node follows whatever CD filesystem the ROM actually carries.

One subtlety worth noting: the scan cannot run at resident init, because the filesystem modules register at lower residentpri than cd.device inits. The boot node registration therefore moved onto the unit task, placed before the disc probe - dosboot turned out to be unhappy about DOS-booting a node that first appears once it is already sitting at the media-wait screen. Re-verified the CD boot end to end under emulation, both with a disc at power-on and with an empty tray.

Kalamatee (approved):

PR #1022, opened 2026-08-19
m68k-amiga: survive a chip-RAM-starved boot console

LinuxJedi:
Follow-up to #1018. With CD32 discs booting, the next thing a commercial title hits is this, looping forever at packet rate:

```
[AmigaVideo:Bitmap] AmigaVideoBM__Root__New: superclass failed to instantiate a suitable bitmap...!!
```

Decoding the tags shows it is not the game's screen at all: 640x256 depth-2 PAL:HighRes with no friend bitmap is the Workbench screen (NominalDimensions hardcodes depth 2), dragged in lazily by the boot CON: window. The game has filled the 2 MB of a stock CD32 by then, so the screen's two contiguous 20 KB chip plane allocations can never succeed - and MakeSureWinIsOpen re-attempted the entire screen open on every ACTION_READ/WRITE/WAIT_CHAR, forever. Making it worse, every one of the eight distinct failure exits in the bitmap class chain produces that identical log line, so the cause was invisible in the field.

Three small commits:

* The planar bitmap class logs (in debug builds) which plane allocation failed, how big it was, and with what MEMF flags - it previously had no failure logging at all.
* The amigavideo failure message, which is already unconditional, now prints AvailMem(MEMF_CHIP) and the largest contiguous chunk beside it. From the failing machine: `chip free 41576, largest 21608` - a one-line diagnosis when the screen needs two contiguous 20 KB planes.
* A boot console that cannot get its window latches the failure and degrades to a sink: writes are swallowed, reads return end-of-file, WaitForChar never sees a character. Programs that only write status text keep running and the machine stops thrashing. Non-boot consoles keep the old per-request behaviour.

Verified under emulation on the CD32 profile: a chip-only machine booting Pinball Fantasies now makes exactly two window attempts (one per console handle) and stops, with the memory line telling the story; with 8 MB of fast RAM the same disc boots through to the game running attract mode; the no-media boot screen path is unchanged.

### DIFF
--- a/arch/m68k-amiga/hidd/amigavideo/amigavideo_bitmapclass.c
+++ b/arch/m68k-amiga/hidd/amigavideo/amigavideo_bitmapclass.c
@@ -11,6 +11,7 @@
 #include <stdio.h>
 #include <string.h>
 
+#include <proto/exec.h>
 #include <proto/oop.h>
 #include <proto/utility.h>
 #include <exec/alerts.h>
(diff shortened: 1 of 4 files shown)


Kalamatee (approved):

LinuxJedi:
Good catch from the review, picked up in 89a6f60e: MakeConWindow's console.device failure path closed the window it had just opened but left fh->window pointing at it, so the next request would treat the freed window as open - handing console I/O a dangling pointer and, with the new latch, bypassing it entirely for that failure shape. The pointer now clears with the close. That bug predates this series; the latch just gave it a second way to hurt. Re-ran the same matrix under emulation: fast-RAM CD boot to the running game, chip-only bounded at two attempts, no-media boot screen unchanged.

PR #1028, opened 2026-08-19
exec: Fix mismatched list lock and Permit in OpenLibrary/OpenDevice

metaneutrons:
Use EXEC_UNLOCK_LIST_AND_PERMIT instead of EXEC_UNLOCK_LIST to properly decrement the Forbid counter initiated by EXEC_LOCK_LIST_READ_AND_FORBID. Without this, tasks were left in an unintended Forbid state, leading to multitasking stalls and potential deadlocks on SMP systems.

### DIFF
--- a/rom/exec/opendevice.c
+++ b/rom/exec/opendevice.c
@@ -97,7 +97,7 @@
 
     iORequest->io_Device = (struct Device *)FindName(&SysBase->DeviceList, devName);
 
-    EXEC_UNLOCK_LIST(&SysBase->DeviceList);
+    EXEC_UNLOCK_LIST_AND_PERMIT(&SysBase->DeviceList);
 
     D(bug("[OpenDevice] Found resident 0x%p\n", iORequest->io_Device));
 
--- a/rom/exec/openlibrary.c
+++ b/rom/exec/openlibrary.c
@@ -67,7 +67,7 @@
     /* Look for the library in our list */
     library = (struct Library *) FindName (&SysBase->LibList, libName);
 
-    EXEC_UNLOCK_LIST(&SysBase->LibList);
+    EXEC_UNLOCK_LIST_AND_PERMIT(&SysBase->LibList);
 
     /* Something found ? */
     if(library!=NULL)


Kalamatee:
Thanks for looking at this - Forbid-count mismatches are nasty to track down, and worth checking. I don't think there's a mismatch here, though, and I think this change would introduce one.

OpenLibrary() and OpenDevice() are balanced already; the pairing is just split across the function rather than contained in one macro:

```
EXEC_LOCK_LIST_READ_AND_FORBID(&SysBase->LibList);   /* Forbid() + spinlock */
library = FindName(&SysBase->LibList, libName);
EXEC_UNLOCK_LIST(&SysBase->LibList);                 /* spinlock only       */
    ... call the library's Open vector ...
Permit();                                            /* balances the Forbid */

```
This holds in both configurations:

- SMP - EXEC_LOCK_LIST_READ_AND_FORBID calls ObtainSystemLock(list, SPINLOCK_MODE_READ, LOCKF_FORBID), which does Forbid() and then takes the spinlock. EXEC_UNLOCK_LIST calls ReleaseSystemLock(list, 0), and with flags == 0 the release path skips its if (flags & LOCKF_FORBID) Permit();, so only the spinlock is dropped and the Forbid is retained.
- non-SMP - the macros reduce to Forbid() and a no-op respectively (rom/exec/exec_locks.h).

Either way there is exactly one outstanding Forbid(), and the explicit Permit() at the end of the function closes it. There are also no early return/goto paths between the lock and that Permit() in either function, so there's no leak on an error path.

With the patch applied, the unlock becomes EXEC_UNLOCK_LIST_AND_PERMIT but the trailing Permit() stays, which gives two Permit()s for one Forbid() on every OpenLibrary()/OpenDevice() call and underflows TDNestCnt.

The second effect is more subtle: the Forbid deliberately spans the Open vector call, not just the list search. Releasing it right after FindName() leaves a window in which another task can expunge or remove the library/device before we call through library / iORequest->io_Device - reachable on SMP in particular.

PR #1034, opened 2026-08-20
m68k: cut the resident chip RAM footprint

LinuxJedi:
While chasing why CD32 discs that run chip-only under Kickstart needed expansion RAM under AROS, I measured where the RAM actually goes on an unexpanded machine. At the point where dosboot hands off to a bootable disc's startup-sequence, Kickstart has around 180KB of the 2MB allocated. AROS had 1.93MB, leaving the game 164KB.

Three things dominated, fixed here in order:

1. **CDVDFS's cache.** `de_NumBuffers = 32` on the CD32 boot node becomes 32 x 16 x 2048 bytes, exactly 1MB, allocated the moment CD0: mounts. With no expansion fitted, MEMF_24BITDMA is chip RAM. Now 4 chunks (128KB).

2. **Pools.** 45 pools are alive after boot, and CreatePool() committed a full puddle to each up front: 323KB of puddles, 313KB of it never touched. The pool header now lives in a block of its own size and the first real puddle arrives with the first AllocPooled(), sized to the request. The handle ABI is unchanged and managed memory keeps the old path. The same 45 pools now cost 7.8KB.

3. **Task stacks.** NewCreateTaskA() floors every stack at AROS_STACKSIZE (16KB on m68k), so the dozen resident tasks pin about 200KB while ten of them peak under 1KB (stacks are MEMF_CLEAR, so the high-water mark is measurable). It also means the TASKTAG_STACKSIZE values already in the tree were being silently ignored. On m68k interrupts run on the supervisor stack, and AmigaOS never floored CreateTask() at all, so the m68k floor drops to 4KB and the resident tasks get explicit sizes with 4-8x margin over the measurements. The CLI default command stack (what a booted game runs on; 4000 bytes on a Kickstart shell) goes from 16KB to 8KB.

Results, measured on stock machines under Copperline, no expansion:

|                                          | before | after  |
|------------------------------------------|--------|--------|
| CD32, free chip at CD boot handoff       | 164KB  | 1.4MB  |
| A1200, free chip at floppy boot handoff  | 1.24MB | 1.64MB |
| A500 (1MB), OS footprint at insert screen | 898KB | 518KB  |

Pinball Fantasies now cold boots to table select on a chip-only CD32, and still does with fast RAM attached. A 512K+512K A500 goes from 64 bytes of free slow RAM to 236KB.

Everything m68k-specific is behind `#ifdef __mc68000` or in m68k-amiga files; other architectures are untouched, and tasks created without an explicit stack size still get the 16KB default. One honest caveat: I have not soak-tested heavy Workbench sessions or the guru display path at the new stack sizes, which is why the alert task deliberately keeps 8KB.

### DIFF
--- a/rom/devs/ata/ata.h
+++ b/rom/devs/ata/ata.h
@@ -38,7 +38,16 @@
 
 #define MAX_DEVICEBUSES         2
 #define MAX_BUSUNITS            2
+/*
+ * The daemon and bus tasks run plain message/IO loops; measured peak
+ * stack use on m68k is a few hundred bytes, and every resident task's
+ * stack comes out of chip RAM on an unexpanded Amiga.
+ */
+#ifdef __mc68000
+#define STACK_SIZE              4096
+#else
 #define STACK_SIZE              16384
+#endif
 #define TASK_PRI                10
 #define TIMEOUT                 30
 
--- a/arch/m68k-amiga/card/card_init.c
+++ b/arch/m68k-amiga/card/card_init.c
@@ -123,6 +123,13 @@ static int Cardres_Init(struct CardResource *CardResource)
         TASKTAG_PC, CardTask,
         TASKTAG_NAME, CardResource->crb_LibNode.lib_Node.ln_Name,
         TASKTAG_PRI, 15,
+        /*
+         * The card task waits for status-change signals and calls the
+         * registered client hooks; measured peak stack use is a few
+         * hundred bytes, and its stack is chip RAM on an unexpanded
+         * machine.
+         */
+        TASKTAG_STACKSIZE, 4096,
         TASKTAG_ARG1, FindTask(0),
         TASKTAG_ARG2, CardResource,
         TAG_DONE);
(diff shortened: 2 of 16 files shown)


Kalamatee:
My concern is that if any of these things are usurped by an interrupt, it may use too much stack now and crash AROS.

rom/devs/ata/ata.h
Kalamatee, line 1: This breaks the ISO-8859-1 copyright symbol (the aros copyright hasnt been updated on touched files in general either).

LinuxJedi:
Interrupts can't land on these stacks on m68k: the CPU switches to the supervisor stack on every exception, and the Amiga port swaps in a dedicated 8K system stack early in boot (arch/m68k-amiga/boot/start.c). The whole server/handler chain, softints included, runs there in supervisor mode and returns via rte (arch/m68k-amiga/kernel/amiga_irq.c), the same way Kickstart's exec does. A task stack only ever holds user-mode frames, which is how classic machines got away with CreateTask's 4K default and games running on 1-2K stacks. It's also why everything here is gated on __mc68000 - on ports where interrupts do run on the current task's stack, the 16K floor is untouched.

The sizes weren't guessed either: task stacks are MEMF_CLEAR, so I measured each resident task's high-water mark after boot plus some exercising, and the peaks came in between roughly 300 and 950 bytes. The new sizes keep 4-10x headroom over that, and anything the scheduler or an interrupt had ever pushed onto those stacks would have shown up in the measurement.

Good catch on ata.h - that was my editor mangling the ISO-8859-1 copyright symbol as a side effect of the edit, not a deliberate change. I've restored the original byte and force-pushed, so the PR no longer touches that line or any of the copyright headers.

Kalamatee (approved):

PR #1055, opened 2026-08-23
usb2otg: declare memset for inline helpers

bitplane:
Failed compile in my build system


bitplane:
Actually, there's more of these. I'll do some more digging.

bitplane:
Yeah this is the wrong fix, it's in FD_ZERO.

PR #1056, opened 2026-08-23
GCC 16 / AArch64: memset is no longer implicit

bitplane:
Updating to GCC 16 for the AArch64 build exposed an implicit import:

* GCC 6.5 tolerated undeclared memset.
* GCC 6.5’s AArch64 runtime lacked floating-point helpers.
* Moving to GCC 16.2 supplied the helpers.
* Fails compile in GCC 16


Kalamatee:
Please do not merge master into your branches - rebase them properly

image

I cannot accept these since they keep messing up the history of the repository.

Kalamatee (approved):

PR #1057, opened 2026-08-23
Bugfix/aarch64 gcc16 bootstrap

bitplane:
3 issues found while building for aarch64 in gcc 16:

### GCC rejected conflicting section attributes

Separate startup pointers from .aros.init section.

### missing aarch64_icache_invalidate() in boot.h

Just declare it

### GCC ignores __attribute__((optnone))

Use optimize("O0") for bootstrap memory routines in GCC, or we might get optimized.
I think this caused an early boot crash but changed other things at the same time. It's safer anyway.

### DIFF
--- a/arch/aarch64-native/kernel/kernel_startup.c
+++ b/arch/aarch64-native/kernel/kernel_startup.c
@@ -77,13 +77,8 @@ asm (
     ".string \"Native/CORE AArch64 v1 (" __DATE__ ")\"" "\n\t\n\t"
 );
 
-#if defined(__clang__)
-static uint64_t * const stack_end __attribute__((used, section(".aros.init"))) = &stack[AROS_STACKSIZE - sizeof(IPTR)];
-static uint64_t * const stack_super_end __attribute__((used, section(".aros.init"))) = &stack_super[AROS_STACKSIZE - sizeof(IPTR)];
-#else
-static uint64_t * const stack_end __attribute__((used, section(".aros.init " TARGET_SECTION_COMMENT))) = &stack[AROS_STACKSIZE - sizeof(IPTR)];
-static uint64_t * const stack_super_end __attribute__((used, section(".aros.init " TARGET_SECTION_COMMENT))) = &stack_super[AROS_STACKSIZE - sizeof(IPTR)];
-#endif
+static uint64_t * const stack_end __attribute__((used, section(".aros.init.data"))) = &stack[AROS_STACKSIZE - sizeof(IPTR)];
+static uint64_t * const stack_super_end __attribute__((used, section(".aros.init.data"))) = &stack_super[AROS_STACKSIZE - sizeof(IPTR)];
 
 struct ARM_Implementation __arm_arosintern  __attribute__((aligned(8), section(".data"))) = {0,0,NULL,0};
 struct ExecBase *SysBase __attribute__((section(".data"))) = NULL;
--- a/arch/aarch64-native/kernel/mmakefile.src
+++ b/arch/aarch64-native/kernel/mmakefile.src
@@ -74,8 +74,6 @@ $(OSGENDIR)/boot/core.elf: $(KOBJSDIR)/kernel_resource.o $(KOBJSDIR)/exec_librar
 #MM kernel-kernel-raspi-aarch64 : includes
 
 USER_INCLUDES := $(PRIV_KERNEL_INCLUDES) -I$(GENINCDIR) -I$(AROS_DEVELOPER)/include -I$(SRCDIR)/rom/openfirmware
-USER_CPPFLAGS := -DTARGET_SECTION_COMMENT=\"$(AROS_SECTION_COMMENT)\"
-
 %build_archspecific \
   mainmmake=kernel-kernel modname=kernel maindir=rom/kernel arch=raspi-aarch64 \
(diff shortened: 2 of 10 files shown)


Kalamatee:
I went through this with the aarch64 clang 11 and arm-aros-gcc 9.5.0 crosstools. The section change is right. A few follow-ups below, and one part of the support.c change I think should be reconsidered.

## The `.aros.init.data` change is correct

For the record on why the warning happens: the entry `asm()` declares `.aros.init` as `"ax"`, and the pointers then go into that same section name as data (`"aw"`). That is a flag conflict inside one translation unit, which is what produces `Warning: ignoring changed section attributes for .aros.init`.

Measured results:

| section spec | gcc / GNU as | clang |
|---|---|---|
| `.aros.init` (current clang branch) | warns | no warning |
| `.aros.init //` (current gcc branch) | no warning | separate, literally named section |
| `.aros.init.data` (this PR) | no warning | no warning |

I also linked both variants against a script replicating `*(SORT_NONE(.aros.init*))`. The output is identical: `.text` comes out `AX`, `start` is still at offset 0, and the new input section is `A` (read only, because the pointers are `const`), so it does not make `.text` writable.

One thing worth putting in the commit message: the `TARGET_SECTION_COMMENT` mechanism is fragile for reasons beyond the clang issue. configure picks the comment character by only checking that a test program links. On ARM, `#` links fine but actually produces a separate section literally named `.aros.init#`, and that only works because the linker script globs. Removing the trick here is a genuine improvement, not just a clang workaround.

## Please make the same change in `arch/arm-native/kernel/kernel_startup.c`

It carries the identical `#if defined(__clang__)` split for `stack_end`, `stack_super_end` and `stack_fiq_end`, and its linker script globs the same way (`*(SORT_NONE(.aros.init*))`), so the same change applies unmodified. Without it, ARM keeps the fragile path while aarch64 is fixed.

Once that is done, `-DTARGET_SECTION_COMMENT=...` is unused in both kernel `mmakefile.src` files (nothing else in those directories references it), so it can be dropped as well.

## Worth a warning comment: do not copy this into the boot code

This technique is safe here because the kernel linker scripts glob. The boot scripts do not. `arch/aarch64-raspi/boot/ldscript.lds:17`, `arch/arm-raspi/boot/ldscript.lds:9` and `arch/riscv-native/sifive_u/boot/ldscript.lds:9` all use an exact `*(.aros.startup)`. If the same rename were applied to `tmp_stack_ptr` in `boot.c`, which the comment there says sits at a fixed location reused by the reset handler, it would be silently orphaned unless the script is changed to `*(SORT_NONE(.aros.startup*))` at the same time.

Note also that `arch/aarch64-raspi/boot/boot.c` cannot produce this warning in the first place, because its `.aros.startup, "ax"` declaration lives in `startup64.S`, a different file. The diagnostic is per translation unit.

## support.c: `-fno-builtin` is already there, and is not sufficient

`arch/aarch64-raspi/boot/mmakefile.src:8` already has `USER_CFLAGS := -Wall $(CFLAGS_NO_BUILTIN)`, and `CFLAGS_NO_BUILTIN` expands to `-fno-builtin`. So that flag was in place while the problem was happening.

What I measured on the actual `memset` body, clang 11, aarch64, `-O2`:

| flags | NEON emitted | recursive call |
|---|---|---|
| `-fno-builtin` | `stp q0, q0` | none |
| `-fno-builtin -mstrict-align` | more NEON, not less | none |
| `-fno-builtin -fno-vectorize` | none | none |

So `-fno-builtin` deals with the recursion half but not the vectorisation half. clang still turns the byte loop into a 128 bit `stp q0, q0`, which is exactly the access the comment says faults with the MMU off. `-mstrict-align` does not help either, since the vectorised stores are aligned and it has no objection to them.

The premise of the change is therefore correct, but I do not think `optnone` / `optimize("O0")` is the right instrument:

1. GCC's manual says of the `optimize` attribute that it "should be used for debugging purposes only. It is not suitable in production code." It resets the whole option set for the function.
2. It costs a lot of code quality. Text size for `memset` at clang `-O2` is 0x108 (264 bytes) with `optnone` against 0x68 (104 bytes) with `-fno-vectorize`, and `optnone` spills everything to the stack. The bootstrap copies a multi megabyte kernel image through these routines.
3. It only protects the three functions that remember to carry the attribute, whereas the whole bootstrap runs with the MMU off.

On the gcc side I confirmed the same picture with arm-aros-gcc 9.5.0: `-O2` does not vectorise, `-O3` does, `optimize("O0")` suppresses it at 0x14c, and `-fno-tree-vectorize` suppresses it at 0xfc.

Suggestion: state the constraint once at build level instead, for example

```make
USER_CFLAGS := -Wall $(CFLAGS_NO_BUILTIN) $(CFLAGS_NO_VECTORIZE)
```

with `CFLAGS_NO_VECTORIZE` configured as `-fno-vectorize` for clang and `-fno-tree-vectorize` for gcc, next to the existing `CFLAGS_NO_BUILTIN*` entries in `config/compiler.cfg.in`. That covers all bootstrap code rather than three functions, and keeps `-O2` quality. If you would rather keep a per function attribute, `__attribute__((optimize("no-tree-vectorize")))` is a much safer choice than `optimize("O0")`.

Kalamatee (approved):

PR #1071, opened 2026-08-27
sdcard fixes

bsek:
## sdcard: error handling, 64-bit safety and transfer fixes

Eight small fixes to sdcard.device, found while bringing the Pi 4 up on aarch64.

- **IPTR for MMIO** — the BCM byte/word accessors truncated the controller
  base address on 64bit builds.
- **Command errors latched separately** — `sdcb_BusStatus` is overwritten by
  the next interrupt, so an error could be erased before the waiting task saw
  it, or a stale one inherited by a command that never failed. The reset in
  the interrupt handler now also drops the listeners, since neither response
  nor data survives it.
- **CMD7 failure is fatal to unit start** — a card left in stand-by answers
  CMD9 but ignores CMD17/CMD18, which previously only surfaced as a command
  timeout much later.
- **Bus flags preserved** — registering a unit overwrote what the bus had
  established about DMA and interrupt delivery.
- **Bounce before PIO** — a buffer too fragmented or out of 32bit reach
  dropped the whole chunk to PIO instead of retrying through the bounce
  buffer.
- **Direct PIO data port reads** — the indirect accessor cost about as much
  as the bus access itself. Writes keep it for the inter-write delay.
- **Aborted chunks not counted** — an unacknowledged CMD12 leaves the card
  streaming, so the data neither arrived intact nor reached the card.
- Controller/clock/bus-width setup logged under `DINIT`.

### DIFF
--- a/rom/devs/sdcard/sdcard_bus.c
+++ b/rom/devs/sdcard/sdcard_bus.c
@@ -1166,9 +1166,30 @@ ULONG FNAME_SDCBUS(FinishData)(struct TagItem *DataTags, struct sdcard_Bus *bus)
                 DTRANS(bug("[SDBus%02u] %s: Attempting to %s %dbytes\n", bus->sdcb_BusNum, __PRETTY_FUNCTION__, ((sdDataMode == MMC_DATA_READ) ? "read" : "write"), tranlen));
                 if (sdDataMode == MMC_DATA_READ)
                 {
-                    for (currword = 0; currword < tranwords; currword++)
+                    /*
+                     * The data port is one 32bit register, so this is a bus
+                     * round trip per word whatever we do - but the indirect
+                     * accessor call costs about as much as the access itself,
+                     * so read through a local pointer, several at a time.
+                     * Writes keep the accessor: the earlier controller needs
+                     * its inter-write delay, and that lives in there.
+                     */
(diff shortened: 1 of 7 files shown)


rom/devs/sdcard/sdcard_bus.c
```diff
@@ -1114,9 +1166,30 @@ ULONG FNAME_SDCBUS(FinishData)(struct TagItem *DataTags, struct sdcard_Bus *bus)
...
+                        dataPtr[currword + 2] = AROS_LE2LONG(*port);
+                        dataPtr[currword + 3] = AROS_LE2LONG(*port);
+                        dataPtr[currword + 4] = AROS_LE2LONG(*port);
+                        dataPtr[currword + 5] = AROS_LE2LONG(*port);
+                        dataPtr[currword + 6] = AROS_LE2LONG(*port);
+                        dataPtr[currword + 7] = AROS_LE2LONG(*port);
+                    }
+                    for (; currword < tranwords; currword++)
                     {
-                        dataPtr[currword] = bus->sdcb_IOReadLong(SDHCI_BUFFER, bus);
```
Kalamatee, line 1119: The generic code should _always_ go through the bus provided functions - not try to take shotcuts like this.

Kalamatee (approved):

PR #1094, opened 2026-08-30
task: derive per-task CPU usage on arm

bsek:
arm has no periodic sweep maintaining iet_CpuUsage, so compute it when TaskTag_CPUUsage is read, from how fast the task's busy time grows.


rom/task/QueryTaskTagList.c
Kalamatee, line 1: This should not be getting done like this in the generic code.

Kalamatee (approved):

PR #1095, opened 2026-08-30
graphics/diskfont: arbitrate the font list with a semaphore

bsek:
Forbid no longer excludes other cores, so the TextFonts walkers take the graphics font semaphore instead. Export it as a private call for diskfont.library, which walks the same list.


rom/graphics/graphics.conf
Kalamatee, line 1: This shouldnt be modifying the amigaos graphic library's API - it needs to be handled another way.

bsek: Reworked: the semaphore is now a public named semaphore ("graphics.library/TextFonts") registered with AddSemaphore(), and diskfont.library looks it up with FindSemaphore() at init. No change to the graphics.library function table.

Kalamatee (approved):

PR #1109, opened 2026-09-01
cd32: fix Microcosm CDXL playback and reduce CD32 chip RAM usage

warpdesign:
Fixes Microcosm and other memory-sensitive CD32 games by:

- Correcting CDXL startup, completion, and zero-length transfer handling.
- Calling CDXL callbacks through the documented register ABI while preserving C nonvolatile registers.
- Implementing the required SystemControlA() controls, including input and low-level library handling.
- Reducing CD32 boot-time chip RAM usage from 1,236,496 to 1,562,168 bytes free.
- Restarting the CIA timer correctly after an aborted request.

Tested with Microcosm, Pinball Illusions, and Superfrog on a stock CD32 configuration with 2 MB chip RAM:

Capture d’écran 2026-09-01 à 01 50 32

Capture d’écran 2026-09-01 à 01 51 23

Capture d’écran 2026-09-01 à 10 06 49

Capture d’écran 2026-09-01 à 10 03 54

### DIFF
--- a/arch/m68k-amiga/nvdisk/akiko.c
+++ b/arch/m68k-amiga/nvdisk/akiko.c
@@ -1,0 +1,644 @@
+/*
+    Copyright (C) 2026, The AROS Development Team. All rights reserved.
+
+    CD32 Akiko 24C08 nonvolatile storage backend.
+*/
+
+#include "nvdisk_intern.h"
+#include "nvdisk_arch.h"
+
+#include <exec/memory.h>
+#include <exec/semaphores.h>
+#include <proto/exec.h>
+
+#include <string.h>
+
+#define AKIKO_ID_ADDR       (*(volatile UWORD *)0x00b80002)
+#define AKIKO_ID_MAGIC      0xcafe
+#define AKIKO_I2C_LINES     (*(volatile UBYTE *)0x00b80030)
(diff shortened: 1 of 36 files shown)


Kalamatee (approved):

arch/m68k-amiga/nvdisk/akiko.c
Kalamatee, line 1: This belongs in the amiga-m68k build, not in generic nvdisk — otherwise every other platform carries code it can never use.

rom/dos/createnewproc.c
Kalamatee, line 1: Please don't add #ifdef __mc68000 to generic code — we've been removing these, not adding more.

Put the value in an arch header and keep a single fallback at the top of this file:

#ifndef PROC_STACKSIZE
#define PROC_STACKSIZE AROS_STACKSIZE
#endif

then just use PROC_STACKSIZE in the code below. That's what AROS_STACKSIZE was for.

Also __mc68000 is true for any m68k target, not just amiga — so this override needs to come from the amigaos-m68k build, not a CPU test.

Kalamatee (changes requested):
see the individual comments

warpdesign:
Should be ok now.

Kalamatee (approved):

PR #1121, opened 2026-09-04
camdusbmidi: build the class on arm and aarch64

bsek:
The arm rule was disabled in e4b16b1210 because the CAMD driver blob would not link. It was also the wrong artefact: the class writes it to DEVS:Midi/ for camd.library to LoadSeg(), and OpenMidiDevice() only accepts a driver whose MidiDeviceData.Init points inside the loaded seglist, so the relocations have to survive. %rule_link_binary flattened it with --oformat binary, which drops every relocation and which the ARM and AArch64 BFD backends refuse outright. The object is self-contained, so embed it as it is. MidiDeviceData needs __attribute__((used)): only camd's MDD_Magic scan refers to it, so the compiler had dropped it.

Compile-tested on arm and aarch64

### DIFF
--- a/rom/usb/classes/camdmidi/camd/mmakefile.src
+++ b/rom/usb/classes/camdmidi/camd/mmakefile.src
@@ -6,7 +6,6 @@ CAMDOBJDIR := $(OBJDIR)/../camdusbmidi/arch
 
 SRCFILE := poseidonusb
 
-USER_LDFLAGS := -static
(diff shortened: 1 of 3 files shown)


Kalamatee (commented):
How this works needs reviewed - The device cannot be written when booting on read only media, or write protected filesystems.

Kalamatee (approved):

PR #1123, opened 2026-09-04
Certificates: update ca-bundle.crt to the Mozilla root store of 2026-08-13

tomaszstaniak:
`ENV:SYS/Certificates/ca-bundle.crt` was extracted from Mozilla's certdata.txt on 2018-04-22. Since then Mozilla has distrusted and removed 79 of those 133 roots (all GeoTrust, thawte and VeriSign roots, the TrustCor roots, and others) and added 67, among them the roots most current sites chain to, such as ISRG Root X2 and GTS Root R1/R3/R4. A TLS client trusting only the old file fails verification against a growing share of the web.

Replaced with curl's `cacert.pem` as published at https://curl.se/ca/cacert.pem (Mozilla data as of 2026-08-13, 121 roots, sha256 `f66dff1bdf8f96060b8177976f8b7d9254bc89bc4db933d769f7384d28480bc9`, matching the published checksum). It is the same extraction the old file came from. The file is used by curl through `CURL_CA_BUNDLE` in `Curl-Startup` and by any OpenSSL client pointed at `ENV:SYS/Certificates`.


Kalamatee:
It may be better to look into automating the retrieval/build of this at build time.

Kalamatee (approved):

PR #1131, opened 2026-09-05
rp1 usb

bsek:
Enables USB on raspi 5 (tested on 500+)


arch/aarch64-raspi/usb/pcixhci/make.opts
Kalamatee, line 1: this should be done in the mmakefile's using an arch specific override.

rom/usb/pcixhci/pcixhci.h
Kalamatee, line 1: The hc->hc_Flags & HCF_PLATFORM behaviour should be getting handled in the platforms PCI subsystem, not directly in the drivers.

bsek:
@Kalamatee  

`HCF_PLATFORM` and `make.opts` replaced by `pcirp1.hidd`

Kalamatee (approved):

PR #1158, opened 2026-09-07
InstallAROS: enumerate destination drives

cyberman:
**Companion translation PR:**

https://github.com/aros-translation-team/installaros/pull/4

---

The destination-drive selector now uses drives actually found by InstallAROS
instead of exposing an editable static device list. Each normal-mode entry keeps
the device and unit together and shows model/capacity where available.
`usbscsi.device` units are probed separately because they are outside the normal
HIDD storage hierarchy.

For targets that cannot be enumerated, **Show developer options** exposes the
manual Device/Unit pair plus **Test Target**. Entering and leaving that section
keeps the detected target coherent. Target checks report open, unit and
read-only errors before an install action uses a manually entered target.
Partition-library validation is action-specific instead of being imposed on
every target.

The partitioning page separates **Target**, **Installation mode**, and
**Developer options** into native Zune groups.

Built for pc-i386 without compiler warnings or errors. Runtime checks covered
normal target selection, developer-option entry/exit, valid and invalid
Device/Unit checks, restoration of the detected target, and the updated native
Zune layout on a disposable QEMU disk.

### DIFF
--- a/workbench/tools/InstallAROS/ia_bootloader.c
+++ b/workbench/tools/InstallAROS/ia_bootloader.c
@@ -44,7 +44,6 @@
 
 extern char *source_Path;
 
-extern Object   *cycle_drivetype;
 extern Object   *optObjCycleGrub2Mode;
 BOOL            gfx_font_exists;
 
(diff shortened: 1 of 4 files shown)


Kalamatee:
This behaviour is not desired. The fields are editable, because not all device drivers support AROS's storage subsystem - and changing them this way prevents users' ability to manually specify the device/unit combination.

cyberman:
Yes, preserving manual device/unit entry makes sense for drivers outside the Storage HIDD. 

My concern is that this shouldn't have to be the normal installation path. InstallAROS is also the first contact for new users, and for detected drives it can present a recognizable target without requiring knowledge of AROS device names.

I think the better split is detected-drive selection by default, with explicit manual device/unit entry as an expert path. Both can still resolve to the same device/unit target internally.

cyberman:
> This behaviour is not desired. The fields are editable, because not all device drivers support AROS's storage subsystem - and changing them this way prevents users' ability to manually specify the device/unit combination.

Rework is done.

PR #1166, opened 2026-09-08
 m68k: Fix stack alignment and minimum process stack size

warpdesign:
Fix m68k process-stack allocation by enforcing the required stack alignment and an 8 KiB minimum stack size.

This prevents stack corruption and crashes affecting programs that open CON: windows. The larger minimum stack slightly reduces available memory on constrained systems; further memory-footprint reductions will be handled separately.

### DIFF
--- a/arch/m68k-all/include/aros/cpu.h
+++ b/arch/m68k-all/include/aros/cpu.h
@@ -24,6 +24,7 @@
 #define AROS_IPTRALIGN             2 /* Alignment for IPTR */
 #define AROS_DOUBLEALIGN           2 /* Alignment for double */
 #define AROS_WORSTALIGN            4 /* Worst case alignment */
+#define AROS_STACKALIGN            8 /* Required for aligned(8) stack objects */
 
 #define AROS_SLOWSTACKFORMAT 1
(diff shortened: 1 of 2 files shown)


arch/m68k-all/include/aros/cpu.h
```diff
@@ -24,6 +24,7 @@
 #define AROS_IPTRALIGN             2 /* Alignment for IPTR */
 #define AROS_DOUBLEALIGN           2 /* Alignment for double */
 #define AROS_WORSTALIGN            4 /* Worst case alignment */
+#define AROS_STACKALIGN            2 /* The stack pointer must remain even */
```
codewiz: I believe the process stack must start aligned to 8 bytes for things like `__attribute__((aligned(8))` to work on local variables.

I just fixed m68k-amigaos-gcc 16.2 to lower the maximum object alignment from 1024 bytes (an big lie) to 8 bytes (a promise that both AllocMem() and LoadSeg() can keep):

https://github.com/AmigaPorts/gcc/pull/36

Kalamatee (approved):

PR #1175, opened 2026-09-10
nvme fixes

bsek:
Three independent fixes to nvme.device, found bringing up an NVMe disk on a Pi 500+.

- **LBA range descriptor pointer arithmetic**: the cast bound to the buffer
  rather than the sum, so `+ 4096` scaled by the 64-byte descriptor size and
  the controller wrote 256KB past the buffer. Affects any controller that
  answers Get Features for LBA Range Type.
- **Fall back to the namespace size when no LBA range is reported**: LBA Range
  Type is optional; a controller with no ranges defined answers successfully
  with an empty descriptor, which gave a zero-length namespace and no unit.
- **Do not wait on a flag the queue task clears**: the drain loop waited on
  `ceh_Reply`, which the queue task clears again after replying. Losing that
  race spun the count out on a finished request. It now also watches the slot
  ceasing to be ours, which does not come back.

### DIFF
--- a/rom/devs/nvme/nvme_busclass.c
+++ b/rom/devs/nvme/nvme_busclass.c
@@ -538,12 +538,16 @@ BOOL Hidd_NVMEBus_Start(OOP_Object *o, struct NVMEBase *NVMEBase)
                         bug ("[NVME:Bus] NVMEBus_Start: ns#%u           nlb %08x%08x\n", nn + 1, (rt->nlb >> 32), rt->nlb & 0xFFFFFFFF);
                     )
 
-                    if (!(rt->attributes & NVME_LBART_ATTRIB_HIDE)) {
+                    if (rt->attributes & NVME_LBART_ATTRIB_HIDE) {
+                        D(bug ("[NVME:Bus] NVMEBus_Start:      Skipping hidden namespace\n");)
+                        lbaEnd = 0;
+                    } else if (rt->nlb) {
                         lbaStart = rt->slba;
                         lbaEnd =  rt->slba + rt->nlb;
                     } else {
-                        D(bug ("[NVME:Bus] NVMEBus_Start:      Skipping hidden namespace\n");)
-                        lbaEnd = 0;
+                        /* Optional feature - no ranges defined still reports success. */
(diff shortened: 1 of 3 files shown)


rom/devs/nvme/nvme_busclass.c
```diff
@@ -509,7 +509,8 @@ BOOL Hidd_NVMEBus_Start(OOP_Object *o, struct NVMEBase *NVMEBase)
                 for (i = 0; i < id_ns->nlbaf + 1; i++) {
                     D(bug ("[NVME:Bus] NVMEBus_Start: ns#%u       lbaf[%u], ms = %u, ds = %u, rp = %u\n", nn + 1, i, id_ns->lbaf[i].ms, id_ns->lbaf[i].ds, id_ns->lbaf[i].rp);)
                 }
-                struct nvme_lba_range_type *rt = (struct nvme_lba_range_type *)(IPTR)buffer + 4096;
+                /* Second page of the buffer - casting first would scale the offset. */
+                struct nvme_lba_range_type *rt = (struct nvme_lba_range_type *)((UBYTE *)buffer + 4096);
```
Kalamatee: using IPTR is the preferred way on AROS - there is no reason to change it to (UBYTE *) here, or in other places.

rom/devs/nvme/nvme_init.c
```diff
@@ -329,13 +329,9 @@ static int NVME_Probe(struct NVMEBase *NVMEBase)
     return TRUE;
 }
 
-/*
- * nvme.device main code has two init routines with 0 and 127 priorities.
- * All bus scanners must run between them.
- */
+/* The set is ordered by priority as text, so init at 0 runs before the scan. */
```
Kalamatee: "init routines run in ascending numeric priority, so NVME_Init at 0 runs before NVME_Probe at 30." Would be the correct comment here - It is not "ordered by priority as text". collect-aros parses the suffix with strtol and inserts entries in ascending numeric order, in tools/collect-aros/gensets.c. Text ordering would put 127 before 30, which is exactly the bug the new comment implies does not exist.

Kalamatee (approved):

PR #1177, opened 2026-09-10
tools/package: normalize extracted paths safely

cyberman:
Root-level PKG members fail to extract under Python 3 because `os.path.dirname()` returns `b''`, while the extractor compares it against `''`.

Review also uncovered the more general cause: the on-disk pathname field is padded to a 4-byte boundary, and `readFile()` currently exposes that padding to callers. This breaks nested members as well as root-level members whenever the pathname length requires padding.

Normalize the pathname at the format boundary in `readFile()` using C-string semantics, so both root and nested members receive the actual pathname.

Doing that also makes previously padded absolute or `../` member names usable pathnames instead of failing incidentally on embedded NULs. Therefore extraction now resolves each normalized member against the destination root and rejects absolute paths, parent traversal, and escapes through pre-existing symlinks before creating directories or files.

Verified against the current upstream master and an AROS-generated `aros-bsp.pkg`:

* package creation remains byte-identical
* root and nested extraction succeeds
* all four pathname alignment classes are covered
* all 24 historical BSP members extract with matching payload/cursor behavior
* absolute and parent-relative paths are rejected
* pre-existing symlink escapes are rejected
* apply/reverse/recovery checks pass

This also subsumes the pathname-containment work from #1178.

### DIFF
--- a/tools/package/pkg
+++ b/tools/package/pkg
@@ -54,8 +54,7 @@ def readFile( file ):
         file.read( struct.calcsize( PKG_FILE_PATHLENGTH_FORMAT ) )
     )
     
-    path = file.read( pathLength[0] )
-    file.read(1)
+    path = file.read( pathLength[0] + 1 ).split( b'\0', 1 )[0]
     
     dataLength = struct.unpack(
         PKG_FILE_DATALENGTH_FORMAT,
@@ -100,6 +99,7 @@ def unpack( source, destination ):
     
     oldwd = os.getcwd()
     os.chdir( destination )
+    destinationRoot = os.path.realpath( os.getcwdb() )
     
     version = readHeader( input )
     
@@ -108,8 +108,16 @@ def unpack( source, destination ):
         
         print('« %s (%d bytes)' % (path, len( data )))
         
+        target = os.path.realpath( os.path.abspath( path ) )
+        try:
+            contained = os.path.commonpath((destinationRoot, target)) == destinationRoot
+        except ValueError:
+            contained = False
+        if os.path.isabs( path ) or not contained:
+            raise ValueError('Package member path escapes destination: %r' % path)
+
         directory = os.path.dirname( path )
-        if directory != '' and not os.path.exists( directory ):
+        if directory and not os.path.exists( directory ):
             os.makedirs( directory )
             
         output = open( path, 'wb' )


Kalamatee:
Thanks for tracking this down - the `!= ''` vs `b''` comparison is the right diagnosis for the root-level failure. The NUL-padding fix is placed in the wrong spot though, so this only repairs the case `aros-bsp.pkg` can exercise.

**Nested members still fail.** `writeFile` packs every name into `(len + 4) & ~3` bytes and stores `pathLength` as that minus one, so any name whose length is not 3 mod 4 comes back from `readFile` with trailing NULs - nested or not. With the strip only in the `if not directory:` branch, a package built from a directory (e.g. `AROS.pkg` from `arch/i386-pc/boot/floppy/mmakefile.src`, which contains `S/Startup-Sequence`) takes the `elif`, `makedirs` succeeds, and then `open(b'S/Startup-Sequence\x00', 'wb')` raises `ValueError: embedded null byte`. Extraction stops with a partial tree. I reproduced this with the script from this branch on a package it built itself.

**Suggested fix.** The padding belongs to the on-disk path field, so strip it once in `readFile` with C-string semantics - the same thing `bootstrap.c` and `workbench/c/Unpack/package.c` do - and `unpack()` does not need a special case at all:

```python
# readFile: replace the two reads with one C-string read
path = file.read( pathLength[0] + 1 ).split( b'\0', 1 )[0]
```

```python
# unpack: plain truthiness test, no strip needed here
directory = os.path.dirname( path )
if directory and not os.path.exists( directory ):
    os.makedirs( directory )
```

That covers root and nested members, keeps `pack` output byte-identical, and also makes the progress line print the real filename (currently it prints the padded `bytes` repr before the strip, so the logged name and the file written differ).

**Pre-existing, not blocking** - noting for a possible follow-up:
- `unpack()` does not reject absolute or `../` member names, so an old package with absolute paths (pre-263f799ef2) extracts outside the destination.
- `os.chdir( oldwd )` is not in a `try/finally`, so any exception mid-extract leaves the cwd inside the destination.
- The input handle in `unpack()` is never closed (`python3 -X dev` warns).

cyberman:
Thanks — agreed on the nested case and on "readFile()" being the right format boundary for the padding.

I hit one interaction while testing the generic fix though: normalizing the pathname before extraction can widen the pre-existing traversal issue you mention. A padded "../" or absolute member that currently dies with "embedded null byte" can become a usable pathname once the C-string padding is removed.

That is why I kept this PR limited to the root-level failure and split pathname containment out before enabling generic nested normalization.

Once containment is in place, I agree that the final padding fix should use C-string semantics in "readFile()" rather than carrying the padding into "unpack()". I'll rework the nested follow-up in that direction.

The cwd restoration and unclosed input handle are good catches as well; I'll keep those separate.

cyberman:
> Thanks for tracking this down - the `!= ''` vs `b''` comparison is the right diagnosis for the root-level failure. The NUL-padding fix is placed in the wrong spot though, so this only repairs the case `aros-bsp.pkg` can exercise.
> 
> **Nested members still fail.** `writeFile` packs every name into `(len + 4) & ~3` bytes and stores `pathLength` as that minus one, so any name whose length is not 3 mod 4 comes back from `readFile` with trailing NULs - nested or not. With the strip only in the `if not directory:` branch, a package built from a directory (e.g. `AROS.pkg` from `arch/i386-pc/boot/floppy/mmakefile.src`, which contains `S/Startup-Sequence`) takes the `elif`, `makedirs` succeeds, and then `open(b'S/Startup-Sequence\x00', 'wb')` raises `ValueError: embedded null byte`. Extraction stops with a partial tree. I reproduced this with the script from this branch on a package it built itself.
> 
> **Suggested fix.** The padding belongs to the on-disk path field, so strip it once in `readFile` with C-string semantics - the same thing `bootstrap.c` and `workbench/c/Unpack/package.c` do - and `unpack()` does not need a special case at all:
> 
> ```python
> # readFile: replace the two reads with one C-string read
> path = file.read( pathLength[0] + 1 ).split( b'\0', 1 )[0]
> ```
> 
> ```python
> # unpack: plain truthiness test, no strip needed here
> directory = os.path.dirname( path )
> if directory and not os.path.exists( directory ):
>     os.makedirs( directory )
> ```
> 
> That covers root and nested members, keeps `pack` output byte-identical, and also makes the progress line print the real filename (currently it prints the padded `bytes` repr before the strip, so the logged name and the file written differ).
> 
> **Pre-existing, not blocking** - noting for a possible follow-up:
> 
>     * `unpack()` does not reject absolute or `../` member names, so an old package with absolute paths (pre-263f799ef2) extracts outside the destination.
> 
>     * `os.chdir( oldwd )` is not in a `try/finally`, so any exception mid-extract leaves the cwd inside the destination.
> 
>     * The input handle in `unpack()` is never closed (`python3 -X dev` warns).

Thanks — I reworked the PR around your `readFile()` suggestion.

The pathname is now normalized once at the format boundary using C-string semantics, so the root-level special case is gone and nested members are covered as well.

This now also subsumes #1178.

Kalamatee (approved):

PR #1178, opened 2026-09-10
 tools/package: contain extracted member paths

cyberman:
Package member paths are currently used directly during extraction, allowing absolute or parent-relative paths to escape the selected destination.

Resolve each member path against the extraction root and reject paths that resolve outside it before creating directories or writing the file. This also prevents escapes through pre-existing symlinks that resolve outside the destination.

Valid root-level and nested relative paths remain unchanged, and package creation output is unaffected.


cyberman:
Superseded by #1177

PR #1188, opened 2026-09-13
repository: document LF line endings for C/H source

cyberman:
AROS already uses Unix line endings for C/H source through its coding conventions and `.astyle` configuration.

Document that expectation explicitly in `CONTRIBUTING.md` and add a minimal `.editorconfig` so supporting editors can follow it automatically.

`CONTRIBUTING.md` also notes that the repository's `.astyle` configuration can mangle MUI UI source layouts, so these need special care when tidying source.

This does not change Git's existing line-ending behaviour or introduce repository-side line-ending conversion.


Kalamatee:
This appears to change the repository's existing “as-is” line-ending behaviour rather than simply documenting the preferred C/H line ending. text=auto eol=lf causes Git to classify these files as text and enforce LF in the working tree, which is precisely the sort of automatic conversion the repository has historically avoided because of the potential for unintended changes/corruption.

If the intent here is only to document that C/H source should normally use LF (e.g. to match astyle), could we keep that in .editorconfig and avoid adding the text=auto eol=lf attributes? If there is a specific Git-related problem this is intended to solve, it would be useful to document that, along with why changing the existing as-is policy is considered safe.

cyberman:
That makes sense. My intent is not to change the repository's existing as-is behaviour, but to make the existing LF convention for C/H source more visible to contributors and easier to follow.

I'll drop the `.gitattributes` part and keep the EditorConfig setting. I'll also add a short note to `CONTRIBUTING.md` so the expectation is documented explicitly rather than relying only on the coding conventions and astyle configuration.

Kalamatee:
> I'll also add a short note to `CONTRIBUTING.md` so the expectation is documented explicitly rather than relying only on the coding conventions and astyle configuration.

please also note the astyle config causes "MUI" UI source layouts to get mangled, so special care is needed to make sure they don't get malformed when tidying files.

cyberman:
> > I'll also add a short note to `CONTRIBUTING.md` so the expectation is documented explicitly rather than relying only on the coding conventions and astyle configuration.
> 
> please also note the astyle config causes "MUI" UI source layouts to get mangled, so special care is needed to make sure they don't get malformed when tidying files.

Good point. I've added that warning to `CONTRIBUTING.md` as well, so contributors are explicitly reminded to check MUI UI source layouts when tidying files with astyle.

Kalamatee (approved):

PR #1206, opened 2026-09-14
Fix three Zune attribute-handling bugs

bsek:
- Scrollbutton VertProp getter returned horiz_prop instead of vert_prop. 
- Process dispatcher's OM_GET case called Process__OM_SET instead of Process__OM_GET. 
- Boopsi's MUIA_Boopsi_MaxHeight handler (in both OM_NEW and OM_SET) wrote to boopsi_minwidth instead of boopsi_maxheight.


workbench/libs/muimaster/classes/process.c
Kalamatee, line 1: is OM_SET not being handled at all?

Kalamatee (approved):

PR #1211, opened 2026-09-14
build: make default build prerequisites explicit

cyberman:
Record GNU Make 4.0 as the minimum supported version and check it during configure.

Document the recursive submodule checkout, canonical toolchain defaults and the normal pc-i386 build sequence. Also document the Autoconf generator contract for the checked-in configure script.


CONTRIBUTING.md
Kalamatee: I'm not sure those requirements need added to the contributing guidelines

cyberman (commented):
Should work

Kalamatee:
> Should work

The point is, those requirements are not related to contributing to the project - so they shouldn't be in that file. They aren't really general rules.

PR #1216, opened 2026-09-15
Fix memory leak by freeing unused affinity masks

bsek:


Kalamatee:
I haven't looked into this in great detail yet so ignore me if I have this wrong - but I think we want each task to have its own mask, based on its parents, and not sharing them, so they can be manipulated individually at runtime if desired.

bsek:
Your comment is not spot on, as this PR fixes another problem with orphaned masks. But, either way it needs a revisit so closing this.

PR #1218, opened 2026-09-15
Add flexgroup layout engine to Zune

bsek:
**Flexgroup**: a Group that lays its children out like a CSS flexbox.

The class owns a MUIA_Group_LayoutHook and drives the standard
MUILM_MINMAX / MUILM_LAYOUT protocol from it, so no change to 
Group itself.
    
Implemented functionality: direction, gap, justify-content, align-items, 
flex-grow, flex-shrink and flex-basis. flex-wrap is not implemented.


Kalamatee:
This should probably be a stand alone class in workbench/classes/zune

bsek:
Class moved as requested

Kalamatee (approved):

PR #1229, opened 2026-09-16
wifi prefs and cli tool

bsek:
An attempt at making it easier to to configure wireless networks in AROS. Brings up a window where you can scan, connect  and disconnect to your preferred network. It displays your assigned ip address and which network you are connected to. 

Currently gated behind raspi targets: `S2INFO_Disassociate` is new and only implemented in bwfm. prism2 would need just that tag, the other wireless drivers lack scanning entirely.


Kalamatee:
This looks to make some assumptions about there only being one wifi interface in a system (which can be incorrect).

bsek:
Fair point. The device chooser does list every device that answers `S2_GETNETWORKS`, and scan/status/details work per selected device — but the connect path assumes a single radio in three inherited ways: `StopSupplicant()` finds WirelessManager by task name, `Wireless.prefs` is a single file hardcoded in `main_amiga.c`, and `AROSTCP/WirelessDevice` names one device.

Supporting several radios properly means one supplicant instance and one config file per device, which is a WirelessManager change rather than a prefs-program one. I can do that in a follow-up PR.

I have updated this PR to handle several radios per device.

Kalamatee (approved):

PR #1246, opened 2026-09-19
Gorilla: use AROS-Shell title for Shell launcher

cyberman:
The Gorilla Shell launcher overrides the `AROS-Shell` window title used by both `CLI` and `NewShell` with `AmigaShell`.

Use the existing `AROS-Shell` title consistently in the Medium and Small Gorilla Shell icon metadata.

Verified both generated icon variants with `ilbmtoicon`/`infoinfo`. Runtime-tested on pc-i386 in QEMU:

* existing carrier Shell: `AmigaShell`
* Gorilla Medium candidate: `AROS-Shell`
* Gorilla Small candidate: `AROS-Shell`

No Shell behavior or window geometry is changed.


Kalamatee:
need confirmation this wasn't done on purpose (for compatibility with AmigaOS shells).

cyberman:
I can't positively confirm the original intent behind the `AmigaShell` title, but there is an additional point which is relevant here: `AROS-Shell` is already the native default used by AROS itself.

`workbench/system/CLI.c` uses:

`CON:0/50//130/AROS-Shell/CLOSE`

and `workbench/c/shellcommands/NewShell.c` uses:

`CON:10/50/640/480/AROS-Shell/CLOSE`

So the Gorilla Shell icon is currently overriding the title used by both native entry paths with `AmigaShell`.

I also traced the icon history:

* `d7973d8a` (2012) introduced the Gorilla `System/Shell.info` with `AmigaShell`. The commit describes it as a “Placeholder icon for System/Shell.info” and does not mention an AmigaOS compatibility requirement.
* `4466d084` (2013) changed the icon from `TYPE = TOOL` to `TYPE = PROJECT`, explicitly noting “same as on OS3.x”. The window title was left unchanged.
* `2258bbf9` (2020) added GorillaSmall and copied the existing metadata.

A repository-wide search does not show any code depending on the `AmigaShell` window title. The only other occurrence is an unrelated comment in `system.c`.

So I can't prove that the 2012 title was not intentional, but I found no documented compatibility contract for it, while the actual AROS CLI and NewShell defaults explicitly use `AROS-Shell`.

If there is a compatibility requirement tied specifically to the Gorilla launcher using `AmigaShell`, please point me to it and I'll rework or withdraw the change accordingly.

PR #1274, opened 2026-09-20
genctbl: allocate input and output paths dynamically

cyberman:
`genctbl` builds the `UnicodeData.txt` input path and generated `.ctbl` / C output paths in fixed 512-byte buffers. Long source or target directories can therefore silently select a truncated input filename or create truncated output filenames.

Allocate those paths to their required size instead of using fixed buffers.

The touched `genctbl.c` source is also normalized from CRLF to LF, matching the repository contribution convention.

Reproduced the input truncation at a 496-byte Unicode directory and the output truncation boundaries at 491 bytes for `.ctbl` output and 494 bytes for C output. With the change, all boundary cases and 2048-byte source/target directories use the exact requested paths.


Kalamatee (approved):

Kalamatee:
Please push CRLF changes separately in future please, otherwise it is hard to review other changes in the commit.

